PR #56, opened 2024-02-27
Fix composition deletion race

jveski:
This fixes the most annoying test flake 🔥 

It's possible for a deleted composition to be updated after its resource slices have been deleted. Currently that will result in a deadlock because the lifecycle controller only updates the current synthesis's observed generation until the slices have been deleted.

If the observed generation doesn't change, the reconstitution cache won't be updated.

This check actually just doesn't make any sense and isn't necessary since the reconstitution cache waits until all resource slices have been deleted before filling anyway.


AYM1607 (approved):
🔥

PR #59, opened 2024-02-29
Fix slice cleanup at scale

jveski:
It's currently possible to outpace the slice cleanup since it's blocked by _any_ in-flight syntheses - even for slices that have long expired.

### DIFF
--- a/internal/controllers/synthesis/slicecleanup.go
+++ b/internal/controllers/synthesis/slicecleanup.go
@@ -92,8 +92,7 @@ func shouldDelete(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool {
 		return false // stale informer
 	}
 	isReferenced := synthesisReferencesSlice(comp.Status.CurrentState, slice) || synthesisReferencesSlice(comp.Status.PreviousState, slice)
-	isSynthesizing := comp.Status.CurrentState == nil || !comp.Status.CurrentState.Synthesized || (comp.Status.PreviousState != nil && !comp.Status.PreviousState.Synthesized)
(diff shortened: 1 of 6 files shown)


internal/controllers/synthesis/slicecleanup.go
```diff
@@ -92,8 +92,7 @@ func shouldDelete(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool {
 		return false // stale informer
 	}
 	isReferenced := synthesisReferencesSlice(comp.Status.CurrentState, slice) || synthesisReferencesSlice(comp.Status.PreviousState, slice)
-	isSynthesizing := comp.Status.CurrentState == nil || !comp.Status.CurrentState.Synthesized || (comp.Status.PreviousState != nil && !comp.Status.PreviousState.Synthesized)
-	return comp.DeletionTimestamp != nil || (!isSynthesizing && !isReferenced)
+	return comp.DeletionTimestamp != nil || (!isReferenced && slice.Spec.CompositionGeneration < comp.Generation)
```
AYM1607, line 95: Why not observed generation?

jveski: That would work too, but it isn't necessary to make this logic work. So I opted for this approach to avoid depending on another controller to update the observed generation. Mostly in case there exists a condition where the observed generation can't be increased until old slices have been GC'd (for some reason)

AYM1607: I see, the current generation will eventually be "observed" but at that point it's irrelevant if these exist because they're no longer referenced by either the current or previous syntheses. Makes sense to me!

AYM1607 (approved):
LGTM

PR #60, opened 2024-02-29
Add separate rate limit for resource slice creation

jveski:
Resource slices can be close to 1mb, so writing a lot of them in a short period is dangerous for apiserver and etcd.

This adds a separate rate limit such that slice creation can be throttled separately from the other, less expensive calls.


AYM1607 (approved):
LGTM

PR #61, opened 2024-02-29
Convert some flags to CR props

jveski:
Moves pod timeout and rollout cooldown into properties of the synthesizer CR.


cmd/eno-controller/main.go
```diff
@@ -27,15 +26,10 @@ func main() {
...
-		rolloutCooldown  time.Duration
-		synthesisTimeout time.Duration
-		debugLogging     bool
-		synconf          = &synthesis.Config{}
+		debugLogging bool
+		synconf      = &synthesis.Config{}
 	)
-	flag.DurationVar(&synconf.Timeout, "synthesis-pod-timeout", time.Minute, "Maximum lifespan of synthesizer pods")
 	flag.Float64Var(&synconf.SliceCreationQPS, "slice-creation-qps", 5, "Max QPS for writing synthesized resources into resource slices")
-	flag.DurationVar(&synthesisTimeout, "synthesis-timeout", time.Second*30, "Timeout when executing synthesizer binaries")
```
AYM1607, line 37: Does timeout and synthesis-timeout have different meanings here?

jveski: Yes- we should document that. Timeout is for each exec into the pod, PodTimeout is for the lifespan of the pod itself. So the controller may retry multiple execs in the same pod, but eventually we will also recreate that pod in case there was something wrong with its node.

AYM1607 (approved):
LGTM

PR #62, opened 2024-03-01
Limit concurrency of terminating pods

jveski:
Currently Eno will keep recreating pods as the current ones are superseded. Naturally this is dangerous since it's possible to end up with an unbounded number of synthesizer pods per composition.

This PR reduces the total max pod concurrency of each composition to 2 - surging by one pod allows fast retries with limited risk.

Also fixes a couple of bugs related to pod lifecycle: corrects the timeout when PodTimeout < ExecTimeout, and avoids exec'ing into deleted pods.

### DIFF
--- a/internal/controllers/synthesis/connection.go
+++ b/internal/controllers/synthesis/connection.go
@@ -71,12 +71,7 @@ func (s *SynthesizerPodConnection) Synthesize(ctx context.Context, syn *apiv1.Sy
 		return nil, fmt.Errorf("creating remote command executor: %w", err)
 	}
 
-	// Execs will in most cases block pod termination, so use the min of the exec and pod lifespan timeouts
-	timeout := syn.Spec.ExecTimeout.Duration
-	if pt := syn.Spec.PodTimeout.Duration; pt < timeout {
-		timeout = pt
-	}
-	streamCtx, cancel := context.WithTimeout(ctx, timeout)
+	streamCtx, cancel := context.WithTimeout(ctx, syn.Spec.ExecTimeout.Duration)
 	defer cancel()
 
 	stderr := &bytes.Buffer{}
--- a/internal/controllers/synthesis/connection.go
+++ b/internal/controllers/synthesis/connection.go
@@ -71,7 +71,12 @@ func (s *SynthesizerPodConnection) Synthesize(ctx context.Context, syn *apiv1.Sy
 		return nil, fmt.Errorf("creating remote command executor: %w", err)
 	}
 
-	streamCtx, cancel := context.WithTimeout(ctx, syn.Spec.ExecTimeout.Duration)
+	// Execs will in most cases block pod termination, so use the min of the exec and pod lifespan timeouts
+	timeout := syn.Spec.ExecTimeout.Duration
+	if pt := syn.Spec.PodTimeout.Duration; pt < timeout {
+		timeout = pt
+	}
+	streamCtx, cancel := context.WithTimeout(ctx, timeout)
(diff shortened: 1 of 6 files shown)


internal/controllers/synthesis/connection.go
```diff
@@ -71,7 +71,12 @@ func (s *SynthesizerPodConnection) Synthesize(ctx context.Context, syn *apiv1.Sy
 		return nil, fmt.Errorf("creating remote command executor: %w", err)
 	}
 
-	streamCtx, cancel := context.WithTimeout(ctx, syn.Spec.ExecTimeout.Duration)
+	// Execs will in most cases block pod termination, so use the min of the exec and pod lifespan timeouts
+	timeout := syn.Spec.ExecTimeout.Duration
+	if pt := syn.Spec.PodTimeout.Duration; pt < timeout {
```
AYM1607, line 76: Can this be somehow specified with kubebuilder and/or [CEL expressions](https://kubernetes.io/docs/reference/using-api/cel/)?

internal/controllers/synthesis/lifecycle.go
```diff
@@ -160,34 +160,40 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
-
 		if comp.DeletionTimestamp != nil {
 			logger = logger.WithValues("reason", "CompositionDeleted")
 			return logger, &pod, true
 		}
 
+		// Allow a single extra pod to be created while the previous one is terminating
+		// in order to break potential deadlocks while avoiding a thundering herd of pods
+		// TODO: e2e test for this
+		pod := pod
```
AYM1607, line 179: Should the shadowing still be done before the composition deletion check? Technically it's fine because we don't care which one of the pods we're returning, but just want to know what you think

internal/controllers/synthesis/lifecycle.go
```diff
@@ -160,34 +160,40 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
 
-func (c *podLifecycleController) shouldDeletePod(logger logr.Logger, comp *apiv1.Composition, syn *apiv1.Synthesizer, pods *corev1.PodList) (logr.Logger, *corev1.Pod /* exists */, bool) {
+func shouldDeletePod(logger logr.Logger, comp *apiv1.Composition, syn *apiv1.Synthesizer, pods *corev1.PodList) (logr.Logger, *corev1.Pod, bool /* exists */) {
 	if len(pods.Items) == 0 {
 		return logger, nil, false
 	}
 
 	// Only create pods when the previous one is deleting or non-existant
+	var onePodDeleting bool
 	for _, pod := range pods.Items {
```
AYM1607, line 170: Should we check the length of this list? Or document how the loop is supposed to behave? For what I can read, if the loop encounters 2 pods that are being deleted the creation of a new one is allowed, but what if the deletion takes a long time? I don't see how we're bounding the total Pod count.

internal/controllers/synthesis/lifecycle.go
jveski: Good call 👍

internal/controllers/synthesis/lifecycle.go
jveski: There's a comment a bit lower that attempts to provide some context. Essentially we already reduce the concurrency of _non-deleted_ pods to 1, so adding this check effectively limits worst-case concurrency to 2.

AYM1607 (approved):
LGTM!

PR #63, opened 2024-03-01
Add metrics

jveski:


internal/controllers/reconciliation/controller.go
```diff
@@ -136,12 +147,17 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 
 func (c *Controller) reconcileResource(ctx context.Context, prev, resource *reconstitution.Resource, current *unstructured.Unstructured) (bool, error) {
 	logger := logr.FromContextOrDiscard(ctx)
+	start := time.Now()
+	defer func() {
+		reconciliationLatency.Observe(time.Since(start).Seconds())
```
AYM1607, line 152: Nice, I completely forgot that arguments of deferred functions are evaluated immediately

internal/controllers/reconciliation/controller.go
```diff
@@ -88,6 +89,16 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 		prev, _ = c.resourceClient.Get(ctx, compRef, &req.Resource)
```
AYM1607, line 89: Unrelated to this PR, but is it safe to ignore the error here?

internal/reconstitution/api.go
```diff
@@ -99,3 +101,23 @@ func (l *lastSeenMeta) MatchesLastSeen(rv string) bool {
...
 	return l.resourceVersion == rv
 }
+
+type lastReconciledMeta struct {
+	lock           sync.Mutex
+	lastReconciled *time.Time
+}
+
+func (l *lastReconciledMeta) ObserveReconciliation() time.Duration {
+	now := time.Now()
```
AYM1607, line 111: Why not sample now after the lock? Realistically I don't think it makes a difference, just curious.

AYM1607 (approved):
LGTM

internal/controllers/reconciliation/controller.go
jveski: Yes because we assume that `prev` can be a nil pointer later on. Some folks think this is a bad practice, I prefer it over passing around a bool. 🤷

internal/reconstitution/api.go
jveski: My split second idea when writing this was to capture any time spent waiting for locks. In retrospect I am not sure how I feel about that, but practically it makes no difference since currently we don't support concurrency in the reconciliation loop.

PR #64, opened 2024-03-01
Fix deletion during synthesis

jveski:
If a synthesis is in-progress when the composition is deleted, there isn't a currentState to clean up - a deadlock state. The fix is to just swap the previous state back to the current.


AYM1607 (approved):
LGTM

PR #65, opened 2024-03-04
Patch the resource slice status instead of updating

jveski:
We can avoid the overhead of deep copying the entire resource slice by using jsonpatch to update only the modified items. This approach also naturally simplifies the signature of the write buffer since the same func can be used for checking the current state and building the patch.

The perf overhead of full updates isn't a big deal now, but will be once we start updating it 2-3x more often (status updates).

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -116,14 +116,15 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 		}
 	}
 
-	c.resourceClient.PatchStatusAsync(ctx, &req.Manifest,
-		func(rs *apiv1.ResourceState) bool {
-			return !((resource.Deleted() && !rs.Deleted) || !rs.Reconciled)
-		},
-		func(rs *apiv1.ResourceState) {
-			rs.Deleted = resource.Deleted()
-			rs.Reconciled = true
-		})
+	c.resourceClient.PatchStatusAsync(ctx, &req.Manifest, func(rs *apiv1.ResourceState) *apiv1.ResourceState {
+		if rs.Deleted == resource.Deleted() && rs.Reconciled {
+			return nil
+		}
+		return &apiv1.ResourceState{
+			Deleted:    resource.Deleted(),
+			Reconciled: true,
+		}
+	})
 
 	if modified {
 		return ctrl.Result{Requeue: true}, nil
--- a/internal/reconstitution/writebuffer.go
+++ b/internal/reconstitution/writebuffer.go
@@ -131,8 +131,9 @@ func (w *writeBuffer) updateSlice(ctx context.Context, sliceNSN types.Namespaced
 	// It's easier to pre-allocate the entire status slice before sending patches
 	// since the "replace" op requires an existing item.
 	if len(slice.Status.Resources) == 0 {
-		slice.Status.Resources = make([]apiv1.ResourceState, len(slice.Spec.Resources))
-		err = w.client.Status().Update(ctx, slice)
+		copy := slice.DeepCopy()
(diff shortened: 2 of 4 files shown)


internal/controllers/reconciliation/controller.go
```diff
@@ -116,14 +116,15 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
...
 
-	c.resourceClient.PatchStatusAsync(ctx, &req.Manifest,
-		func(rs *apiv1.ResourceState) bool {
-			return !((resource.Deleted() && !rs.Deleted) || !rs.Reconciled)
-		},
-		func(rs *apiv1.ResourceState) {
-			rs.Deleted = resource.Deleted()
-			rs.Reconciled = true
-		})
+	c.resourceClient.PatchStatusAsync(ctx, &req.Manifest, func(rs *apiv1.ResourceState) *apiv1.ResourceState {
```
AYM1607, line 119: Did you move to this pattern to favor immutability? Or did something else in the code made this pattern not ergonomic?

internal/reconstitution/writebuffer.go
```diff
@@ -175,3 +176,9 @@ func (w *writeBuffer) updateSlice(ctx context.Context, sliceNSN types.Namespaced
 	logger.V(0).Info(fmt.Sprintf("updated the status of %d resources in slice", len(updates)))
 	return true
 }
+
+type jsonPatch struct {
```
AYM1607, line 180: I feel it coming 👀

internal/reconstitution/writebuffer.go
```diff
@@ -129,44 +128,46 @@ func (w *writeBuffer) updateSlice(ctx context.Context, sliceNSN types.Namespaced
...
-	}
-
-	// The resource slice informer doesn't DeepCopy automatically for perf reasons.
-	// So we can apply the CheckFn to status pointers but not PatchFn.
-	// We'll deep copy the status here only when it needs to change.
+	var patches []*jsonPatch
 	for _, update := range updates {
 		unsafeStatusPtr := &unsafeSlice[update.SlicedResource.Index]
-		if update.CheckFn(unsafeStatusPtr) {
+		patch := update.PatchFn(unsafeStatusPtr)
```
AYM1607, line 149: unsafe, very rusty of you :D

internal/reconstitution/writebuffer.go
```diff
@@ -129,44 +128,46 @@ func (w *writeBuffer) updateSlice(ctx context.Context, sliceNSN types.Namespaced
...
-	if len(slice.Status.Resources) != len(slice.Spec.Resources) {
-		dirty = true
-		status = slice.Status.DeepCopy()
-		status.Resources = make([]apiv1.ResourceState, len(slice.Spec.Resources))
+	// It's easier to pre-allocate the entire status slice before sending patches
+	// since the "replace" op requires an existing item.
+	if len(slice.Status.Resources) == 0 {
+		copy := slice.DeepCopy()
+		copy.Status.Resources = make([]apiv1.ResourceState, len(slice.Spec.Resources))
+		err = w.client.Status().Update(ctx, copy)
```
AYM1607, line 136: Why reset the status on every update?

internal/reconstitution/writebuffer.go
jveski: Go did it first! 🤣 https://pkg.go.dev/unsafe

internal/reconstitution/writebuffer.go
jveski: This is only reached when the status is uninitialized

jveski: Ideally the slice wouldn't need to be initialized before applying patches but other controllers assume the resources slice is either nil or len == len(spec.Resources). So it isn't safe to partially initialize the status currently

AYM1607 (approved):
LGTM

PR #66, opened 2024-03-05
Fix patch idempotence

jveski:
The reconciliation controller isn't actually idempotent currently, although the impact is hidden by the resource version cache.

e.g. if a resource's version hasn't changed, we won't even attempt to build a patch for it. But when we do, we always send it to the server.


AYM1607 (approved):
LGTM

PR #67, opened 2024-03-05
Parse CEL expressions to determine resource readiness

jveski:


internal/controllers/reconciliation/controller.go
```diff
@@ -111,8 +115,23 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
...
 
+	// Check if resource status is ready since we can't use the resource version cache for non-ready resources
+	// (otherwise we wouldn't be able to evaluate their readiness)
+	var ready bool
+	slice := &apiv1.ResourceSlice{}
+	err = c.client.Get(ctx, req.Manifest.Slice, slice)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("getting resource slice: %w", err)
+	}
+	if len(slice.Status.Resources) > req.Manifest.Index {
```
AYM1607, line 126: Can the statuses be non-empty and have a different length than the manifests the slice holds?

internal/controllers/reconciliation/controller.go
```diff
@@ -244,8 +278,8 @@ func (c *Controller) buildPatch(ctx context.Context, prev, resource *reconstitut
 	return patch, types.StrategicMergePatchType, err
 }
 
-func (c *Controller) getCurrent(ctx context.Context, resource *reconstitution.Resource, apiVersion string) (*unstructured.Unstructured, bool, error) {
-	if resource.HasBeenSeen() && !resource.Deleted() {
+func (c *Controller) getCurrent(ctx context.Context, resource *reconstitution.Resource, apiVersion string, disableCache bool) (*unstructured.Unstructured, bool, error) {
```
AYM1607, line 281: Where is the connection between disable cache and !ready coming from?

jveski: We can't use the metadata API to check the resource version when the resource isn't ready, since we wouldn't get the signal required to mark the resource as ready. There are some related comments up higher in the file, lmk if you think there's a more clear way to document this

internal/controllers/reconciliation/controller.go
jveski: Practically, no. But it is technically possible if another client tampers with the status, so I think it's worth handling

jveski: I fixed the logic and now the tests are failing lol. Will investigate in the morning

internal/controllers/reconciliation/controller.go
```diff
@@ -123,7 +123,7 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("getting resource slice: %w", err)
 	}
-	if len(slice.Status.Resources) > req.Manifest.Index {
+	if len(slice.Status.Resources) >= req.Manifest.Index-1 {
```
AYM1607, line 126: I think the original logic was fine? if length == 5 and index == 6 this passes and it shouldn't

jveski: lol you're right. That's what I get for coding during meetings 😬

AYM1607 (approved):
LGTM

PR #68, opened 2024-03-07
Add smoke tests

### LINKED ISSUE #43: E2E testing
The closest thing that we have to an E2E test in the repo is the exec integration test which is currently broken. We should have a real E2E test pipeline at least for the happy path.

jveski:
Resolves #43 

Just a quick smoke test to make sure the example code works, and that we don't totally break the synthesis pod lifecycle code (that can't be easily integration tested).

### DIFF
--- a/cmd/eno-controller/main.go
+++ b/cmd/eno-controller/main.go
@@ -23,14 +23,21 @@ func main() {
 	}
 }
 
+// TODO: Handle case where resource slices are already deleted but comp still references them
+
 func run() error {
 	ctx := ctrl.SetupSignalHandler()
 	var (
 		debugLogging bool
 		synconf      = &synthesis.Config{}
+
+		mgrOpts = &manager.Options{
+			Rest: ctrl.GetConfigOrDie(),
+		}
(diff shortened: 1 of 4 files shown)


cmd/eno-controller/main.go
```diff
@@ -23,14 +23,21 @@ func main() {
 	}
 }
 
+// TODO: Handle case where resource slices are already deleted but comp still references them
```
jveski, line 26: Still need to add a finalizer to synthesizers...

cmd/eno-controller/main.go
```diff
@@ -23,14 +23,21 @@ func main() {
...
+
 func run() error {
 	ctx := ctrl.SetupSignalHandler()
 	var (
 		debugLogging bool
 		synconf      = &synthesis.Config{}
+
+		mgrOpts = &manager.Options{
```
jveski, line 34: Apparently we didn't parse the controller manager flags in this process before

AYM1607 (approved):
LGTM

PR #70, opened 2024-03-08
Aggregate resource readiness signal back to the composition

jveski:


api/v1/composition.go
```diff
@@ -64,5 +64,5 @@ type Synthesis struct {
 
 	Synthesized bool `json:"synthesized,omitempty"`
 	Ready       bool `json:"ready,omitempty"`
-	Synced      bool `json:"synced,omitempty"`
+	Reconciled  bool `json:"reconciled,omitempty"`
```
jveski, line 67: We don't use the term "sync" anywhere else, seems out of place here

internal/controllers/synthesis/lifecycle.go
```diff
@@ -131,17 +126,21 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 		}
 
 		// Remove the finalizer when all pods and slices have been deleted
-		if len(sliceList.Items) > 0 || len(pods.Items) > 0 {
-			logger.V(1).Info(fmt.Sprintf("refusing to remove composition finalizer because %d associated resource slices and %d pods still exist", len(sliceList.Items), len(pods.Items)))
-			return ctrl.Result{}, nil // some resources still exist
+		if comp.Status.CurrentState != nil && (!comp.Status.CurrentState.Reconciled) || comp.Status.CurrentState.ObservedCompositionGeneration != comp.Generation {
+			logger.V(1).Info("refusing to remove composition finalizer because some resources have not yet been cleaned up")
```
jveski, line 130: We can use the aggregated status signal when cleaning up instead of the old resource slice index.

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -13,8 +13,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// TODO: Release previous slices once the current set has been reconciled?
```
jveski, line 16: More complexity than it's worth 🔥

internal/controllers/aggregation/status.go
```diff
@@ -0,0 +1,85 @@
...
+	client client.Client
+}
+
+func NewStatusController(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&apiv1.Composition{}).
+		Owns(&apiv1.ResourceSlice{}).
+		WithLogConstructor(manager.NewLogConstructor(mgr, "statusAggregationController")).
+		Complete(&statusController{
+			client: mgr.GetClient(),
```
AYM1607, line 25: This never needs to reach out to the overlay right? just aggregates from slices to compositions?

internal/controllers/aggregation/status.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		// Status might be lagging behind
+		if len(slice.Status.Resources) == 0 && len(slice.Spec.Resources) > 0 {
+			ready = false
+			reconciled = false
+			break
+		}
+
+		for _, state := range slice.Status.Resources {
```
AYM1607, line 59: Why && with `reconciled` and `ready`? You're just avoiding setting the variables to false repeatedly, but it shouldn't make a difference, right? Maybe I'm missing something

internal/controllers/synthesis/lifecycle.go
```diff
@@ -115,13 +106,17 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 			return ctrl.Result{}, nil
 		}
 
+		// TODO: Is it possible to delete before the first synthesis?
```
AYM1607, line 109: When is the finalizer set? Technically if you delete right after creation, yes, right?

internal/controllers/synthesis/lifecycle.go
AYM1607, line 130: Is this log a bit misleading? Could we shortcircuit on a lagging observed generation?

AYM1607 (approved):
LGTM, comments mostly for my edu

internal/controllers/synthesis/lifecycle.go
```diff
@@ -115,13 +106,17 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
 		// Deletion increments the composition's generation, but the reconstitution cache is only invalidated
 		// when the synthesized generation (from the status) changes, which will never happen because synthesis
 		// is righly disabled for deleted compositions. We break out of this deadlock condition by updating
 		// the status without actually synthesizing.
 		if comp.Status.CurrentState != nil && comp.Status.CurrentState.ObservedCompositionGeneration != comp.Generation {
 			comp.Status.CurrentState.ObservedCompositionGeneration = comp.Generation
-			comp.Status.CurrentState.Synthesized = true // in case the previous synthesis failed
+			comp.Status.CurrentState.Ready = false
+			comp.Status.CurrentState.Reconciled = false
+			comp.Status.CurrentState.Synthesized = true // in case the previous synthesis failed (TODO I don't think this actually works)
```
AYM1607, line 119: Interesting, what is this supposed to do?

internal/controllers/aggregation/status.go
jveski: Yeah exactly 👍

internal/controllers/aggregation/status.go
jveski: You're completely right - this was just some leftovers from a slightly different implementation

internal/controllers/synthesis/lifecycle.go
jveski: I was thinking about if a composition is deleted between the time its finalizer is added and the time its first synthesis has succeeded. But it should be okay since we continue on to remove the finalizer in cases where current state is nil. Will remove this note

internal/controllers/synthesis/lifecycle.go
jveski: Yeah I agree the message is not 100% accurate - will change

internal/controllers/synthesis/lifecycle.go
jveski: I don't remember lol 😆 . I'll follow up on the TODO though and leave more comments

PR #72, opened 2024-03-13
Reduce memory consumption

jveski:
Compositions can contain a ton of resources, which are currently cached in the controller's informers, the reconciler's informers, and the reconciler's reconstitution cache. This change reduces the number of in-memory copies to 1: the reconstitution cache.

- Moves slice cleanup controller to reconciler so the controller no longer starts up a resource slice informer
- Strips the `manifest` property out of resource slices in the reconciler informers
- Gets the full resource slice from a non-caching client when filling the reconstitution cache

### DIFF
--- a/examples/loadtest.yaml
+++ b/examples/loadtest.yaml
@@ -4,18 +4,20 @@ metadata:
   name: load-test-synth
 spec:
   image: docker.io/ubuntu:latest
+  execTimeout: 1m
   command:
   - /bin/bash
   - -c
   - |
       n=3000
+      tr -dc A-Za-z0-9 </dev/urandom | head -c 4096 > rando
       echo -n "{\"apiVersion\":\"config.kubernetes.io/v1\",\"kind\":\"ResourceList\",\"items\":["
       for i in $(seq $n); do
         echo -n "{
           \"apiVersion\": \"v1\",
           \"kind\": \"ConfigMap\",
           \"metadata\": { \"name\": \"test-cm-${i}\", \"namespace\": \"default\", \"annotations\": { \"eno.azure.io/reconcile-interval\": \"30s\" } },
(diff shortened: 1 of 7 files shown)


examples/loadtest.yaml
```diff
@@ -4,18 +4,20 @@ metadata:
...
       n=3000
+      tr -dc A-Za-z0-9 </dev/urandom | head -c 4096 > rando
       echo -n "{\"apiVersion\":\"config.kubernetes.io/v1\",\"kind\":\"ResourceList\",\"items\":["
       for i in $(seq $n); do
         echo -n "{
           \"apiVersion\": \"v1\",
           \"kind\": \"ConfigMap\",
           \"metadata\": { \"name\": \"test-cm-${i}\", \"namespace\": \"default\", \"annotations\": { \"eno.azure.io/reconcile-interval\": \"30s\" } },
-          \"data\": { \"foo\": \"$(tr -dc A-Za-z0-9 </dev/urandom | head -c 4096; echo)\" }
+          \"data\": { \"foo\": \"$(cat rando)\" }
```
jveski, line 20: avoid the cpu overhead of generating a long random string for each resource

internal/manager/manager.go
```diff
@@ -91,16 +103,30 @@ func newMgr(logger logr.Logger, opts *Options, isReconciler bool) (ctrl.Manager,
...
+			Field: fieldSelector,
+		}
 	}
 
-	if !isReconciler {
+	if isReconciler {
 		yespls := true
 		mgrOpts.Cache.ByObject[&apiv1.ResourceSlice{}] = cache.ByObject{
 			UnsafeDisableDeepCopy: &yespls,
+			Transform: func(obj any) (any, error) {
```
AYM1607, line 120: Nice!

AYM1607 (approved):
LGTM

PR #73, opened 2024-03-13
Fix informer race condition

jveski:
This fixes an exceptionally rare test flake in which the status of a resource isn't ever populated. It's caused by the status being flushed out of the write buffer before the corresponding resource slice is cached in the informer. This is possible because the reconstitution cache doesn't read from the informer cache. So it's possible to get a composition change event, fill the cache, and reconcile a resource before the resource slice hits the informer.


jveski:
The trick here is that the not found signal is coming from a PUT instead of an informer read. So it will only return 404 if the resource really doesn't exist unlike the informer which may just be out of sync.

AYM1607 (approved):
Makes total sense, approved!

PR #75, opened 2024-03-14
THE LAST BREAKING CHANGE

jveski:
- Normalizes log levels across the various components
- Renames (Current|Previous)State to Synthesis to be more consistent with other terminology
- Switches status bools to timestamps so we can log how long it took for components to be reconciled and become ready
- Allows CEL expressions to optionally return a condition struct instead of only bools. This lets us grab the state transition time from the condition instead of using our own timestamp.
- Removes inputs in anticipation of new reference API
- Moves synthesis pods into a specified namespace


api/v1/resourceslice.go
```diff
@@ -41,17 +41,9 @@ type ResourceSliceStatus struct {
 }
 
 type ResourceState struct {
-	// True when the resource has been sync'd to the specified manifest.
```
AYM1607, line 44: Is this documented anywhere else?

internal/controllers/aggregation/status.go
```diff
@@ -57,29 +59,52 @@ func (s *statusController) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 
 		for _, state := range slice.Status.Resources {
+			state := state
```
AYM1607, line 62: We can thankfully say goodbye to this idiom in 1.22!

internal/reconstitution/cache.go
```diff
@@ -95,7 +95,7 @@ func (c *cache) Fill(ctx context.Context, comp *apiv1.Composition, synthesis *ap
 	compNSN := types.NamespacedName{Name: comp.Name, Namespace: comp.Namespace}
 	c.synthesesByComposition[compNSN] = append(c.synthesesByComposition[compNSN], synKey.Generation)
 
-	logger.Info("cache filled")
+	logger.V(0).Info("cache filled")
```
AYM1607, line 98: I'm not too familiar with `go-logr` what are the "V" chained methods for?

internal/controllers/synthesis/rollout.go
```diff
@@ -90,14 +90,15 @@ func (c *rolloutController) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		logger.Info("advancing synthesizer rollout process")
+		logger.V(0).Info("advancing synthesizer rollout process")
 		return ctrl.Result{RequeueAfter: syn.Spec.RolloutCooldown.Duration}, nil
 	}
 
 	// Update the status to reflect the completed rollout
 	if syn.Status.CurrentGeneration != syn.Generation {
 		previousTime := syn.Status.LastRolloutTime
 		now := metav1.Now()
+		syn.Status.LastRolloutTime = &now
```
AYM1607, line 101: Was this missing here causing any issues?

internal/controllers/synthesis/exec_test.go
```diff
@@ -5,12 +5,9 @@ import (
 	"time"
```
AYM1607, line 5: Goodbye inputs!

internal/controllers/aggregation/status.go
```diff
@@ -57,29 +59,52 @@ func (s *statusController) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 				ready = false
 			}
+			if state.Ready != nil && (maxReadyTime == nil || maxReadyTime.Before(state.Ready)) {
+				maxReadyTime = state.Ready
+			}
 		}
 	}
 
-	if comp.Status.CurrentState.Reconciled == reconciled && comp.Status.CurrentState.Ready == ready {
+	if (comp.Status.CurrentSynthesis.Reconciled != nil) == reconciled && (comp.Status.CurrentSynthesis.Ready != nil) == ready {
```
AYM1607, line 78: The conditionals are getting trickier and trickier to read 😄 we should probably to a cleanup pass later on

internal/controllers/reconciliation/controller.go
```diff
@@ -330,3 +329,33 @@ func mungePatch(patch []byte, rv string) ([]byte, error) {
 
 	return json.Marshal(patchMap)
 }
+
+// evalReadinessChecks evaluates and prioritizes readiness checks.
+// - Nil is returned when less than all of the checks are ready
+// - If some precise and some inprecise times are given, the precise times are favored
+// - Within precise or non-precise times, the max of that group is always used
```
AYM1607, line 336: Could you elaborate a bit on precise vs non-precise times and how they're used?

AYM1607: I see the code in api.go but I'd like to know how we're using it.

api/v1/resourceslice.go
jveski: These comments are included in the generated CRD yaml files so that's something. But also I think we will need nice hand-written docs later on and they should definitely include this.

internal/controllers/aggregation/status.go
jveski: YESSSS

internal/reconstitution/cache.go
jveski: They just set the verbosity like the old glog/klog libraries did in the google tradition

internal/controllers/aggregation/status.go
jveski: Yeah fully agree. I'd like to refactor a bunch of the controller conditionals into pure funcs and write unit tests against them to capture the intention

internal/controllers/reconciliation/controller.go
jveski: Talked offline - essentially precise timestamps are derived from k8s condition transition times, non-precise are captured from the system time by the Eno controller.

AYM1607 (approved):
LGTM!

PR #76, opened 2024-03-18
Fix test race

jveski:
Switches to k8s optimistic concurrency instead of time.Sleep lol


jveski:
I'm going to merge this myself since it's a test-only one-line change and Mariano is out 😸

PR #77, opened 2024-03-20
Race fixes

jveski:


api/v1/composition.go
```diff
@@ -35,8 +35,9 @@ type CompositionStatus struct {
 
 // Synthesis represents a Synthesizer's specific synthesis of a given Composition.
 type Synthesis struct {
-	ObservedCompositionGeneration int64 `json:"observedCompositionGeneration,omitempty"`
-	ObservedSynthesizerGeneration int64 `json:"observedSynthesizerGeneration,omitempty"`
+	UUID                          string `json:"uuid,omitempty"`
```
jveski, line 38: the uuid helps link pods back to a specific synthesis (e.g. regardless of composition or synthesizer generations)

api/v1/synthesizer.go
```diff
@@ -33,19 +33,19 @@ type SynthesizerSpec struct {
 	// Timeout for each execution of the synthesizer command.
 	//
 	// +kubebuilder:default="10s"
-	ExecTimeout metav1.Duration `json:"execTimeout,omitempty"`
+	ExecTimeout *metav1.Duration `json:"execTimeout,omitempty"`
```
jveski, line 36: Imagine how many times I facepalmed upon realizing that non-ptr durations don't support defaulting

go.mod
```diff
@@ -5,95 +5,97 @@ go 1.21
 toolchain go1.21.5
```
jveski, line 5: when I was convinced I was going crazy I tried bumping controller runtime. Didn't hurt so I just left it in this PR

internal/controllers/aggregation/status.go
```diff
@@ -48,7 +48,7 @@ func (s *statusController) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		slice.Namespace = comp.Namespace
 		err := s.client.Get(ctx, client.ObjectKeyFromObject(slice), slice)
 		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("getting resource slice: %w", err)
+			return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("getting resource slice: %w", err))
```
jveski, line 51: not a race fix exactly, but prevents churning on work queue items where the slice no longer exists

internal/controllers/reconciliation/controller.go
```diff
@@ -139,7 +140,7 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 	slice := &apiv1.ResourceSlice{}
 	err = c.client.Get(ctx, req.Manifest.Slice, slice)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("getting resource slice: %w", err)
+		return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("getting resource slice: %w", err))
```
jveski, line 143: same as the other case - this should have existed from the beginning to avoid leaking work items when slices are deleted while resources remain in cache

internal/controllers/reconciliation/controller.go
```diff
@@ -157,7 +158,7 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 
 	// Store the results
 	c.resourceClient.PatchStatusAsync(ctx, &req.Manifest, func(rs *apiv1.ResourceState) *apiv1.ResourceState {
-		if rs.Deleted == resource.Deleted() && rs.Reconciled && rs.Ready != nil && *rs.Ready == *ready {
+		if rs != nil && rs.Deleted == resource.Deleted() && rs.Reconciled && ptr.Deref(rs.Ready, metav1.Time{}) == ptr.Deref(ready, metav1.Time{}) {
```
jveski, line 161: fixes the comparison logic between times

internal/controllers/reconciliation/integration_test.go
```diff
@@ -174,6 +175,7 @@ func TestCRUD(t *testing.T) {
 			syn := &apiv1.Synthesizer{}
 			syn.Name = "test-syn"
 			syn.Spec.Image = "create"
+			syn.Spec.RolloutCooldown = &metav1.Duration{Duration: time.Millisecond}
```
jveski, line 178: needed now that we have working sane defaults

internal/controllers/reconciliation/integration_test.go
```diff
@@ -472,6 +474,7 @@ func TestReconcileStatus(t *testing.T) {
 	comp.Name = "test-comp"
 	comp.Namespace = "default"
 	require.NoError(t, upstream.Create(ctx, comp))
+	comp.ResourceVersion = "1"
```
jveski, line 477: invalidate the copy of the resource received from the server such that the next write will fail until our informers catch up.

internal/controllers/reconciliation/integration_test.go
```diff
@@ -484,11 +487,16 @@ func TestReconcileStatus(t *testing.T) {
 	require.NoError(t, upstream.Create(ctx, slice))
 
 	now := metav1.Now()
-	comp.Status.CurrentSynthesis = &apiv1.Synthesis{
-		Synthesized:    &now,
-		ResourceSlices: []*apiv1.ResourceSliceRef{{Name: slice.Name}},
-	}
-	require.NoError(t, upstream.Status().Update(ctx, comp))
+	err = retry.RetryOnConflict(testutil.Backoff, func() error {
```
jveski, line 490: missing retry

internal/controllers/reconciliation/integration_test.go
```diff
@@ -612,7 +620,7 @@ func TestMidSynthesisDeletion(t *testing.T) {
 	comp.Finalizers = []string{"eno.azure.io/cleanup"}
 	comp.Spec.Synthesizer.Name = syn.Name
 	require.NoError(t, upstream.Create(ctx, comp))
-	comp.ResourceVersion = "" // forget this version so the next write guarantees sync'd informer
+	comp.ResourceVersion = "1" // forget this version so the next write guarantees sync'd informer
```
jveski, line 623: empty strings aren't valid resource versions, so server returns a 400 instead of 409

internal/controllers/reconciliation/integration_test.go
```diff
@@ -893,8 +903,7 @@ func TestHelmOwnershipTransfer(t *testing.T) {
 
 	// Wait for the composition to be sync'd - it shouldn't delete the resource
 	testutil.Eventually(t, func() bool {
-		upstream.Get(ctx, client.ObjectKeyFromObject(comp), comp)
-		return comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.ObservedCompositionGeneration == comp.Generation && comp.Status.CurrentSynthesis.Reconciled != nil
+		return errors.IsNotFound(upstream.Get(ctx, client.ObjectKeyFromObject(comp), comp))
```
jveski, line 906: the logic was just... wrong before. It was possible for the composition to be deleted before the check

internal/controllers/synthesis/exec.go
```diff
@@ -103,11 +104,8 @@ func (c *execController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 func (c *execController) synthesize(ctx context.Context, syn *apiv1.Synthesizer, comp *apiv1.Composition, pod *corev1.Pod) ([]*apiv1.ResourceSliceRef, error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
-	synctx, done := context.WithTimeout(ctx, syn.Spec.ExecTimeout.Duration)
```
jveski, line 106: This was a duplicate timeout

internal/controllers/synthesis/exec.go
```diff
@@ -135,7 +133,8 @@ func (c *execController) fetchPreviousSlices(ctx context.Context, comp *apiv1.Co
 		slice.Name = ref.Name
 		slice.Namespace = comp.Namespace
 		err := c.client.Get(ctx, client.ObjectKeyFromObject(slice), slice)
-		if errors.IsNotFound(err) {
+		if errors.IsNotFound(err) && comp.Status.PreviousSynthesis.Synthesized != nil && time.Since(comp.Status.PreviousSynthesis.Synthesized.Time) > time.Minute*5 {
```
jveski, line 136: seems reasonable to expect informers to catch up in 5min

internal/controllers/synthesis/lifecycle.go
```diff
@@ -112,12 +118,12 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 		// when the synthesized generation (from the status) changes, which will never happen because synthesis
 		// is righly disabled for deleted compositions. We break out of this deadlock condition by updating
 		// the status without actually synthesizing.
-		if comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.ObservedCompositionGeneration != comp.Generation {
+		if comp.Status.CurrentSynthesis != nil && (comp.Status.CurrentSynthesis.ObservedCompositionGeneration != comp.Generation || comp.Status.CurrentSynthesis.Synthesized == nil) {
```
jveski, line 121: it's possible for the generations to match for reasons other than being set here, but we still need to enter this branch to set the synthesized timestamp.

internal/controllers/synthesis/pod.go
```diff
@@ -17,7 +17,10 @@ func newPod(cfg *Config, scheme *runtime.Scheme, comp *apiv1.Composition, syn *a
 	pod.GenerateName = "synthesis-"
 	pod.Namespace = cfg.PodNamespace
 	pod.Finalizers = []string{"eno.azure.io/cleanup"}
-	pod.Labels = map[string]string{manager.ManagerLabelKey: manager.ManagerLabelValue}
+	pod.Labels = map[string]string{
+		manager.ManagerLabelKey:       manager.ManagerLabelValue,
+		"eno.azure.io/synthesis-uuid": comp.Status.CurrentSynthesis.UUID,
```
jveski, line 22: just makes it easier to query for without relying on an informer index

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -47,6 +49,10 @@ func (c *sliceCleanupController) Reconcile(ctx context.Context, req ctrl.Request
 		comp.Name = owner.Name
 		comp.Namespace = slice.Namespace
 		err = c.client.Get(ctx, client.ObjectKeyFromObject(comp), comp)
+		if errors.IsNotFound(err) && time.Since(slice.CreationTimestamp.Time) < time.Minute {
+			logger.V(1).Info("didn't find a composition for this resource slice - ignoring because resource slice is new so informer may just be stale")
```
jveski, line 53: Worse case we slightly delay cleanup of unused resource slices

internal/reconstitution/reconstituter.go
```diff
@@ -107,8 +107,7 @@ func (r *reconstituter) populateCache(ctx context.Context, comp *apiv1.Compositi
 		slice.Namespace = comp.Namespace
 		err := r.nonCachedReader.Get(ctx, client.ObjectKeyFromObject(&slice), &slice)
 		if err != nil {
-			// TODO: Handle 404?
```
jveski, line 110: what do you know, this mattered 🙃

internal/testutil/testutil.go
```diff
@@ -227,9 +230,17 @@ func (m *Manager) GetCurrentResourceSlices(ctx context.Context) ([]*apiv1.Resour
 	return returns, nil
 }
 
+var Backoff = wait.Backoff{
```
jveski, line 233: a more aggressive version of the defaults

internal/testutil/testutil.go
```diff
@@ -227,9 +230,17 @@ func (m *Manager) GetCurrentResourceSlices(ctx context.Context) ([]*apiv1.Resour
...
+	Duration: 10 * time.Millisecond,
+	Factor:   2.0,
+	Jitter:   0.1,
+	Cap:      time.Minute,
+}
+
 func Eventually(t testing.TB, fn func() bool) {
 	t.Helper()
-	SomewhatEventually(t, time.Second*5, fn)
+	SomewhatEventually(t, time.Second*15, fn)
```
jveski, line 243: This probably isn't necessary now that the fixes have been found, but it helps get signal on future races

api/v1/synthesizer.go
AYM1607: Outrageous!!!

internal/controllers/synthesis/lifecycle.go
```diff
@@ -79,7 +79,7 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 	syn.Name = comp.Spec.Synthesizer.Name
 	err = c.client.Get(ctx, client.ObjectKeyFromObject(syn), syn)
 	if err != nil {
-		return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("getting synthesizer: %w", err))
+		return ctrl.Result{}, fmt.Errorf("getting synthesizer: %w", err)
```
jveski, line 82: we shouldn't ever ignore a composition just because its synthesizer doesn't exist, assuming it has one configured. In this case a stale synth cache would cause synthesis not to happen until the next generation of the composition

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -98,7 +104,7 @@ func shouldReleaseFinalizer(comp *apiv1.Composition, slice *apiv1.ResourceSlice)
 	if comp.Status.CurrentSynthesis != nil && slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration && owner.UID == comp.UID {
 		return false // stale informer
 	}
-	return !resourcesRemain(slice) || (!synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) && !synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice))
+	return comp.DeletionTimestamp != nil && (!resourcesRemain(slice) || (!synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) && !synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice)))
```
jveski, line 107: Checking the composition deletion status ensures that we don't prematurely release the resource slice if it is deleted before the composition (or before we receive the composition deletion watch event).

This can happen if apiserver's GC beats our controller to the deletion which happens although very rarely.

internal/controllers/aggregation/status.go
```diff
@@ -61,7 +61,7 @@ func (s *statusController) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		for _, state := range slice.Status.Resources {
 			state := state
 			// Sync
-			if !state.Reconciled || (comp.DeletionTimestamp != nil && !state.Deleted) {
+			if (comp.DeletionTimestamp != nil && !state.Deleted) || !state.Reconciled {
```
jveski, line 64: This one is bonkers. The order matters because if the composition is deleted, we should _only_ care if it's deleted. The state being reconciled is not adequate.

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -113,12 +119,13 @@ func synthesisReferencesSlice(syn *apiv1.Synthesis, slice *apiv1.ResourceSlice)
...
 
-func resourcesRemain(slice *apiv1.ResourceSlice) bool {
+func resourcesRemain(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool {
 	if len(slice.Status.Resources) == 0 && len(slice.Spec.Resources) > 0 {
 		return true // status is lagging behind
 	}
+	shouldOrphan := comp != nil && comp.Annotations != nil && comp.Annotations["eno.azure.io/deletion-strategy"] == "orphan"
 	for _, state := range slice.Status.Resources {
-		if !state.Deleted {
+		if !state.Deleted && !shouldOrphan {
```
jveski, line 128: We no longer set Deleted in the reconciliation controller when orphaning resources, since that doesn't actually make sense. Other callers need to understand orphaning logic now.

internal/controllers/synthesis/slicecleanup_test.go
```diff
@@ -33,7 +33,8 @@ func TestSliceCleanupControllerOrphanedSlice(t *testing.T) {
 			return err
 		}
 		comp.Status.CurrentSynthesis = &apiv1.Synthesis{
-			Synthesized: ptr.To(metav1.Now()),
+			Synthesized:                   ptr.To(metav1.Now()),
+			ObservedCompositionGeneration: comp.Generation,
```
jveski, line 37: This test was slightly wrong before

internal/controllers/synthesis/lifecycle.go
```diff
@@ -91,13 +94,16 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
+			return ctrl.Result{}, nil
+		}
 		return ctrl.Result{RequeueAfter: syn.Spec.PodTimeout.Duration}, nil
 	}
 
 	if comp.DeletionTimestamp != nil {
 		// If the composition was being synthesized at the time of deletion we need to swap the previous
 		// state back to current. Otherwise we'll get stuck waiting for a synthesis that can't happen.
-		if comp.Status.CurrentSynthesis == nil || comp.Status.CurrentSynthesis.Synthesized == nil {
+		if comp.Status.PreviousSynthesis != nil && (comp.Status.CurrentSynthesis == nil || comp.Status.CurrentSynthesis.Synthesized == nil) {
```
jveski, line 106: I'm not sure this is necessary, but seems more sane. Don't swap back if there isn't anything to swap to.

internal/controllers/aggregation/status.go
```diff
@@ -80,7 +88,7 @@ func (s *statusController) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	now := metav1.Now()
-	if ready {
+	if ready && maxReadyTime != nil {
```
jveski, line 91: It's possible to have zero resources, which means ready would be true with a nil maxReadyTime

internal/controllers/reconciliation/controller.go
```diff
@@ -156,19 +157,20 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 	}
 
 	// Store the results
+	if modified {
```
jveski, line 160: We now re-enter the loop _before_ writing status instead of directly after. This ensures that we don't write status that differs from what we get back on the next pass, so less risk of oscillation.

internal/controllers/reconciliation/controller.go
```diff
@@ -156,19 +157,20 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
...
+		return ctrl.Result{Requeue: true}, nil
+	}
+	deleted := current == nil || current.GetDeletionTimestamp() != nil
 	c.resourceClient.PatchStatusAsync(ctx, &req.Manifest, func(rs *apiv1.ResourceState) *apiv1.ResourceState {
-		if rs.Deleted == resource.Deleted() && rs.Reconciled && rs.Ready != nil && *rs.Ready == *ready {
+		if rs != nil && rs.Deleted == deleted && rs.Reconciled && ptr.Deref(rs.Ready, metav1.Time{}) == ptr.Deref(ready, metav1.Time{}) {
```
jveski, line 165: Deletion state is now the actual deletion state of the resource, not a reflection of if we've successfully reconciled our desired state. The two values diverge when resources are purposefully orphaned, or likely in other cases (at least other bugs).

internal/controllers/reconciliation/controller.go
```diff
@@ -190,7 +192,7 @@ func (c *Controller) reconcileResource(ctx context.Context, comp *apiv1.Composit
 			return false, nil // already deleted - nothing to do
 		}
 		if comp.Annotations["eno.azure.io/deletion-strategy"] == "orphan" {
-			return true, nil // signal that we deleted without actually doing so
+			return false, nil
```
jveski, line 195: We didn't actually do anything, so don't pretend that we modified anything.

internal/controllers/synthesis/lifecycle.go
jveski: Hmm I think this might fix deletion during the initial synthesis, since otherwise we'd keep swapping back to previous for every tick of the loop. Made a note to add a test for this scenario.

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -85,20 +91,18 @@ func (c *sliceCleanupController) Reconcile(ctx context.Context, req ctrl.Request
 }
 
 func shouldDelete(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool {
-	owner := metav1.GetControllerOf(slice)
-	if comp.Status.CurrentSynthesis != nil && slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration && owner.UID == comp.UID {
+	if comp.Status.CurrentSynthesis != nil && slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration {
```
jveski, line 94: Matching the UID isn't necessary, and violates one's expectations in cases where resource slices are somehow recreated

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -85,20 +91,18 @@ func (c *sliceCleanupController) Reconcile(ctx context.Context, req ctrl.Request
...
 }
 
 func shouldReleaseFinalizer(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool {
-	owner := metav1.GetControllerOf(slice)
-	if comp.Status.CurrentSynthesis != nil && slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration && owner.UID == comp.UID {
+	if comp.Status.CurrentSynthesis != nil && slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration {
 		return false // stale informer
 	}
-	return !resourcesRemain(slice) || (!synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) && !synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice))
+	return comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Synthesized != nil && (!resourcesRemain(comp, slice) || (!synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) && !synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice)))
```
jveski, line 105: Never release if synthesis is still in flight since it's possible for the observed generation to backtrack i.e. newer resource slices might still be necessary.

internal/testutil/testutil.go
```diff
@@ -141,7 +159,14 @@ func NewManager(t *testing.T) *Manager {
 			panic(err)
 		}
 	})
-	m.DownstreamRestConfig, err = downstreamEnv.Start()
+	for i := 0; i < 2; i++ {
```
jveski, line 162: Sometimes tests flake because this fails to start. 3 retries seems to be enough.

internal/controllers/synthesis/rollout.go
```diff
@@ -80,7 +84,7 @@ func (c *rolloutController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			return ctrl.Result{}, fmt.Errorf("advancing last rollout time: %w", err)
 		}
 
-		err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		err = retry.RetryOnConflict(testutil.Backoff, func() error {
```
AYM1607, line 87: Should we use the testutil version of the default backoff here?

internal/testutil/testutil.go
```diff
@@ -79,7 +86,8 @@ func NewManager(t *testing.T) *Manager {
 			filepath.Join(root, "api", "v1", "config", "crd"),
 			testCrdDir,
 		},
-		ErrorIfCRDPathMissing: true,
+		ErrorIfCRDPathMissing:    true,
+		AttachControlPlaneOutput: os.Getenv("ACTIONS_RUNNER_DEBUG ") != "" || os.Getenv("ACTIONS_STEP_DEBUG ") != "",
```
AYM1607, line 90: Are the trailing spaces needed in the env var names?

internal/testutil/testutil.go
```diff
@@ -118,8 +135,9 @@ func NewManager(t *testing.T) *Manager {
 	version, _ := strconv.Atoi(os.Getenv("DOWNSTREAM_VERSION_MINOR"))
 
 	downstreamEnv := &envtest.Environment{
-		BinaryAssetsDirectory: dir,
-		ErrorIfCRDPathMissing: true,
+		BinaryAssetsDirectory:    dir,
+		ErrorIfCRDPathMissing:    true,
+		AttachControlPlaneOutput: os.Getenv("ACTIONS_RUNNER_DEBUG ") != "" || os.Getenv("ACTIONS_STEP_DEBUG ") != "",
```
AYM1607, line 140: ditto

internal/testutil/testutil.go
jveski: Nope! Fixed.

AYM1607 (approved):
LGTM

PR #79, opened 2024-03-21
Helm ownership transfer

jveski:
Adds test coverage for taking and giving up resource ownership with Helm, along with one small change to allow purposeful resource orphaning to enable this workflow.

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -126,7 +126,7 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 	// Skip without logging since this is a very hot path
 	var modified bool
 	if hasChanged {
-		modified, err = c.reconcileResource(ctx, prev, resource, current)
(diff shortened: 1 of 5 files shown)


internal/controllers/reconciliation/controller.go
```diff
@@ -189,6 +189,9 @@ func (c *Controller) reconcileResource(ctx context.Context, prev, resource *reco
 		if current == nil || current.GetDeletionTimestamp() != nil {
 			return false, nil // already deleted - nothing to do
 		}
+		if comp.Annotations["eno.azure.io/deletion-strategy"] == "orphan" {
```
AYM1607, line 192: nit: Expose these as consts?

AYM1607: We might need to reference them from the ask side

AYM1607 (approved):
LGTM!

internal/controllers/reconciliation/controller.go
jveski: I have weird opinions about consts lol. Let's chat sometime and hash out how to approach consts across the project :)

PR #80, opened 2024-03-21
Exponential backoff for pod creation

jveski:
If a composition is impossible to synthesize, we should eventually slow down the synthesis loop to avoid the cost of starting/running the pods.


internal/controllers/synthesis/integration_test.go
```diff
@@ -321,3 +321,89 @@ func TestControllerSwitchingSynthesizers(t *testing.T) {
 		assert.NotEqual(t, comp.Status.CurrentSynthesis.ResourceSlices, initialSlices)
 	})
 }
+
+func TestControllerBackoff(t *testing.T) {
```
AYM1607, line 325: Should we split these off as well? Maybe pod_backoff_test?

internal/controllers/synthesis/lifecycle.go
```diff
@@ -185,6 +185,17 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, nil
 	}
 
+	// Back off to avoid constantly re-synthesizing impossible compositions (unlikely but possible)
+	if current := comp.Status.CurrentSynthesis; current != nil && current.Attempts > 0 && current.PodCreation != nil {
+		const base = time.Millisecond * 250
```
AYM1607, line 190: Make configurable?

internal/controllers/synthesis/integration_test.go
jveski: I like it - done

AYM1607 (approved):
LGTM

PR #82, opened 2024-03-26
Refactor readiness checks

jveski:
Just moves code related to the readiness check expression parsing/execution into its own package.


AYM1607 (approved):
LGTM

PR #83, opened 2024-03-26
Add manager tests

jveski:
Adds tests to our controller manager configuration and fixes a few bugs found along the way.

### DIFF
--- a/api/v1/resourceslice.go
+++ b/api/v1/resourceslice.go
@@ -19,7 +19,6 @@ type ResourceSlice struct {
 	Status ResourceSliceStatus `json:"status,omitempty"`
 }
 
-// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 type ResourceSliceSpec struct {
 	CompositionGeneration int64      `json:"compositionGeneration,omitempty"`
 	Resources             []Manifest `json:"resources,omitempty"`
--- a/cmd/eno-controller/main.go
+++ b/cmd/eno-controller/main.go
@@ -44,6 +44,7 @@ func run() error {
 	if synconf.PodNamespace == "" {
 		return fmt.Errorf("a value is required in --synthesizer-pod-namespace or POD_NAMESPACE")
 	}
+	mgrOpts.SynthesizerPodNamespace = synconf.PodNamespace
 
 	zapCfg := zap.NewProductionConfig()
 	if debugLogging {
--- a/examples/loadtest.yaml
+++ b/examples/loadtest.yaml
@@ -9,7 +9,7 @@ spec:
   - /bin/bash
   - -c
   - |
-      n=3000
+      n=12000
       tr -dc A-Za-z0-9 </dev/urandom | head -c 4096 > rando
       echo -n "{\"apiVersion\":\"config.kubernetes.io/v1\",\"kind\":\"ResourceList\",\"items\":["
       for i in $(seq $n); do
@@ -67,7 +67,7 @@ spec:
               timeout="11s"
             fi
 
-            kubectl patch synthesizers load-test-synth --type=merge --patch "{\"spec\":{\"timeout\": \"$timeout\"}}"
+            kubectl patch synthesizers load-test-synth --type=merge --patch "{\"spec\":{\"execTimeout\": \"$timeout\"}}"
             n=$(($n + 1))
-            sleep 10
+            sleep 120
           done
--- a/internal/controllers/synthesis/exec.go
+++ b/internal/controllers/synthesis/exec.go
@@ -20,8 +20,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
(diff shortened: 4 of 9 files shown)


api/v1/resourceslice.go
```diff
@@ -19,7 +19,6 @@ type ResourceSlice struct {
 	Status ResourceSliceStatus `json:"status,omitempty"`
 }
 
-// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
jveski, line 22: This started breaking finalizer updates in my test environment. I have no idea how this isn't breaking the tests, but for now it just isn't worth spending the time thinking about.

examples/loadtest.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   - /bin/bash
   - -c
   - |
-      n=3000
+      n=12000
```
jveski, line 12: 🚀

examples/loadtest.yaml
```diff
@@ -67,7 +67,7 @@ spec:
               timeout="11s"
             fi
 
-            kubectl patch synthesizers load-test-synth --type=merge --patch "{\"spec\":{\"timeout\": \"$timeout\"}}"
+            kubectl patch synthesizers load-test-synth --type=merge --patch "{\"spec\":{\"execTimeout\": \"$timeout\"}}"
```
jveski, line 70: keeping up with a crd change

internal/controllers/synthesis/exec.go
```diff
@@ -20,8 +20,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-// maxSliceJsonBytes is the max sum of a resource slice's manifests. It's set to 1mb, which leaves 512kb of space for the resource's status, encoding overhead, etc.
-const maxSliceJsonBytes = 1024 * 768
+// maxSliceJsonBytes is the max sum of a resource slice's manifests.
+const maxSliceJsonBytes = 1024 * 512
```
jveski, line 24: I caught a few cases where we got some errors back from etcd because resource slices were too large. This may need to be refined more in the future

internal/manager/manager.go
```diff
@@ -102,14 +102,14 @@ func newMgr(logger logr.Logger, opts *Options, isController, isReconciler bool)
...
-	podLabelSelector := labels.SelectorFromSet(labels.Set{ManagerLabelKey: ManagerLabelValue})
-
 	if isController {
-		// We do not honor the configured label selector, because these pods will only ever have labels set by Eno.
-		// But we _do_ honor the field selector since it may reduce the namespace scope, etc.
+		// Only cache pods in the synthesizer pod namespace and owned by this controller
 		mgrOpts.Cache.ByObject[&corev1.Pod{}] = cache.ByObject{
-			Label: podLabelSelector,
-			Field: fieldSelector,
+			Namespaces: map[string]cache.Config{
```
jveski, line 108: Now that synthesizer pods are in the controller's namespace, it doesn't really make sense to filter the cache like we currently are. We can calculate the ns/label selectors without requiring any other configuration.

internal/manager/manager.go
```diff
@@ -122,8 +122,8 @@ func newMgr(logger logr.Logger, opts *Options, isController, isReconciler bool)
 				if !ok {
 					return obj, nil
 				}
-				for _, res := range slice.Spec.Resources {
-					res.Manifest = "" // remove big manifest that we don't need
+				for i := range slice.Spec.Resources {
+					slice.Spec.Resources[i].Manifest = "" // remove big manifest that we don't need
```
jveski, line 126: This just... didn't work 😆

internal/manager/manager.go
```diff
@@ -21,13 +21,19 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	apiv1 "github.com/Azure/eno/api/v1"
 )
 
+func init() {
+	log.SetLogger(zap.New(zap.WriteTo(os.Stdout)))
```
jveski, line 34: Turns out this is needed to log non-test warnings from controller runtime

cmd/eno-controller/main.go
```diff
@@ -44,6 +44,7 @@ func run() error {
 	if synconf.PodNamespace == "" {
 		return fmt.Errorf("a value is required in --synthesizer-pod-namespace or POD_NAMESPACE")
 	}
+	mgrOpts.SynthesizerPodNamespace = synconf.PodNamespace
```
AYM1607, line 47: 😮

internal/controllers/synthesis/exec.go
AYM1607: Interesting, I guess the status grows linearly, do you think that was causing the issue?

internal/manager/manager.go
AYM1607: Ah, good ol' shadowing 😄

AYM1607 (approved):
LGMT!

PR #86, opened 2024-03-27
New inputs runtime

AYM1607:


internal/controllers/synthesis/exec_io.go
```diff
@@ -11,9 +11,44 @@ import (
...
+		bindings[b.Key] = &b
+	}
+	refs := map[string]*apiv1.Ref{}
+	for _, r := range syn.Spec.Refs {
+		r := r
+		refs[r.Key] = &r
+	}
+
+	inputs := []*unstructured.Unstructured{}
+	for key, r := range refs {
```
jveski, line 31: Iterate over the refs slice directly to preserve order?

jveski, line 31: Iterate over the refs slice directly to preserve order?

internal/controllers/synthesis/exec_io.go
```diff
@@ -11,9 +11,44 @@ import (
...
+		b, ok := bindings[key]
+		if !ok {
+			return nil, fmt.Errorf("input %q is referenced, but not bound", key)
```
jveski, line 34: Oooh maybe a validating webhook in the future...

internal/controllers/synthesis/exec_io.go
```diff
@@ -11,9 +11,44 @@ import (
...
+		}
+		input := apiv1.NewInput(key, apiv1.InputResource{
+			Name:      b.Resource.Name,
+			Namespace: b.Resource.Namespace,
```
jveski, line 38: I think we should default the namespace to the composition's namespace. Since typically resource references are contained to the referencing resource's namespace e.g. secrets/configmaps from pods, it seems natural to allow omitting the namespace.

I like that we can reference across namespaces though. It's all too common to get stuck replicating secrets between namespaces just to reference them from pods

examples/inputs/eno.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: eno.azure.io/v1
+kind: Synthesizer
+metadata:
+  name: test-synth
+spec:
+  refs:
+  - key: some-input
+    resource:
+      group: someGroup
```
jveski, line 9: Out of curiosity, does passing any string here work? And does passing an empty string work (as I'd expect for the core group)?

internal/controllers/synthesis/exec_io_test.go
```diff
@@ -0,0 +1,81 @@
+package synthesis
+
+import (
+	"testing"
+
+	apiv1 "github.com/Azure/eno/api/v1"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildPodInput(t *testing.T) {
```
jveski, line 10: Nice unit tests 🥇 

Can you also add a TODO for an integration test to cover inputs? Not worth adding now, but probably a good idea longer term

internal/controllers/synthesis/exec_io.go
AYM1607: Yup, I'll add a TODO

internal/controllers/synthesis/exec_io.go
AYM1607: Discussed offline, we don't want to default because it can cause problems with non-namespaced resources.

examples/inputs/eno.yaml
AYM1607: It does have to be a valid group, this example just showcases that input data is passed correctly to synthesizers. I'll work on a proper example that actually reaches out to apiserver on a follow up PR.

jveski (approved):

PR #88, opened 2024-03-27
3 tiny test fixes

jveski:
- Removes a flaky assertion that isn't useful anyway
- Extends the load test's synthesizer timeout
- Greatly reduces the risk of flakes in another time-based test (I've only seen this one flake once anyway)

### DIFF
--- a/internal/controllers/synthesis/backoff_test.go
+++ b/internal/controllers/synthesis/backoff_test.go
@@ -43,8 +43,8 @@ func TestControllerBackoff(t *testing.T) {
 			return comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Attempts >= 3
 		})
 
-		// It shouldn't be possible to try three times within 1s
-		assert.Greater(t, int(time.Since(start).Milliseconds()), 1000)
+		// It shouldn't be possible to try three times within 250ms
+		assert.Greater(t, int(time.Since(start).Milliseconds()), 250)
 	})
 }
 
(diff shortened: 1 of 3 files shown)


internal/controllers/synthesis/backoff_test.go
```diff
@@ -43,8 +43,8 @@ func TestControllerBackoff(t *testing.T) {
 			return comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Attempts >= 3
 		})
 
-		// It shouldn't be possible to try three times within 1s
-		assert.Greater(t, int(time.Since(start).Milliseconds()), 1000)
+		// It shouldn't be possible to try three times within 250ms
+		assert.Greater(t, int(time.Since(start).Milliseconds()), 250)
```
jveski, line 47: I've seen this test flake once because pod events caused the next controller tick before the incremented counter from the previous attempt was in cache. Reducing this value significantly still proves the controller defers retries but will essentially eliminate the risk of flakes.

AYM1607: > because pod events caused the next controller tick before the incremented counter from the previous attempt was in cache
Is this something inherent to the test setup? Or should it be addressed in the controller?

jveski: Real pods move much slower than our fake test pod control loop, so it's essentially impossible to hit the race in real use. Also it would "fail open" i.e. skip backoff and continue to create a pod anyway.

We could address it in the controller, but doing so would require an extra etcd write

AYM1607 (approved):
🏄‍♂️

PR #90, opened 2024-03-27
Accept Pod ServiceAccount as a flag

AYM1607:


internal/controllers/synthesis/pod.go
```diff
@@ -32,6 +32,7 @@ func newPod(cfg *Config, scheme *runtime.Scheme, comp *apiv1.Composition, syn *a
 	userID := int64(1000)
 	yes := true
 	pod.Spec = corev1.PodSpec{
+		ServiceAccountName: cfg.PodServiceAccount,
```
AYM1607, line 35: Safe to pass an empty string here, we could also default the flag to "default" with the same effect. Any preference?

jveski: I like this approach - the pod resource should own its own defaulting

jveski (approved):
:shipit:

PR #94, opened 2024-03-27
Refactor synthesis controller

jveski:
Moves some code out to a new package, renames a few things for clarity, breaks out some gnarly expressions into separate functions, etc.

### DIFF
--- a/internal/controllers/synthesis/exec.go
+++ b/internal/controllers/synthesis/exec.go
@@ -12,12 +12,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 // maxSliceJsonBytes is the max sum of a resource slice's manifests.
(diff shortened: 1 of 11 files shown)


internal/controllers/synthesis/exec.go
```diff
@@ -137,8 +135,7 @@ func (c *execController) fetchPreviousSlices(ctx context.Context, comp *apiv1.Co
 		slice.Name = ref.Name
 		slice.Namespace = comp.Namespace
 		err := c.noCacheClient.Get(ctx, client.ObjectKeyFromObject(slice), slice)
-		if errors.IsNotFound(err) && comp.Status.PreviousSynthesis.Synthesized != nil && time.Since(comp.Status.PreviousSynthesis.Synthesized.Time) > time.Minute*5 {
```
jveski, line 140: This isn't possible now since we've fixed the manager - we're making real calls to apiserver instead of hitting the informer cache.

internal/controllers/synthesis/lifecycle.go
```diff
@@ -102,56 +102,8 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	if comp.DeletionTimestamp != nil {
-		// If the composition was being synthesized at the time of deletion we need to swap the previous
```
jveski, line 105: To separate func to keep this one short(er)

internal/controllers/synthesis/exec.go
```diff
@@ -151,76 +148,6 @@ func (c *execController) fetchPreviousSlices(ctx context.Context, comp *apiv1.Co
 	return slices, nil
 }
 
-func buildResourceSlices(comp *apiv1.Composition, previous []*apiv1.ResourceSlice, outputs []*unstructured.Unstructured, maxJsonBytes int) ([]*apiv1.ResourceSlice, error) {
```
jveski, line 154: To new package

internal/controllers/synthesis/exec.go
```diff
@@ -287,40 +214,3 @@ func truncateString(str string, length int) (out string) {
 	}
 	return out + "[truncated]"
 }
-
-func consumeReconcileIntervalAnnotation(obj client.Object) *metav1.Duration {
```
jveski, line 291: To new package

internal/resource/resource.go
```diff
@@ -0,0 +1,131 @@
+package resource
```
jveski, line 1: This file deserves another pass of refactoring and some unit tests - will address in a separate PR.

internal/resource/slicing.go
```diff
@@ -0,0 +1,121 @@
+package resource
```
jveski, line 1: What do you think about the naming here? `resource.Slice` and `resource.New`

AYM1607: In general, I like `New` when "main" type of a package is named after the package itself. If that's not the case or there's multiple types the `NewType` format, but that's just personal preference. I don't mind `Slice` :)

jveski: I like it - renamed

AYM1607 (approved):
LGTM

PR #95, opened 2024-03-28
Add watchdog controller

jveski:
kube-state-metrics can't compare timestamps against the current time so we need to implement that monitoring logic within Eno.

### DIFF
--- a/internal/controllers/watchdog/controller.go
+++ b/internal/controllers/watchdog/controller.go
@@ -1,0 +1,71 @@
+package watchdog
+
+import (
+	"context"
+	"time"
+
+	apiv1 "github.com/Azure/eno/api/v1"
+	"github.com/Azure/eno/internal/manager"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// watchdogController exposes metrics that track the states of Eno resources relative to the current time.
+// The idea is to identify deadlock states so they can be alerted on.
+type watchdogController struct {
+	client    client.Client
+	threshold time.Duration
+}
+
+func NewController(mgr ctrl.Manager, threshold time.Duration) error {
+	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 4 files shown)


AYM1607:
> kube-state-metrics can't compare timestamps against the current time

Sadness 😢

internal/controllers/watchdog/controller.go
```diff
@@ -0,0 +1,71 @@
...
+// watchdogController exposes metrics that track the states of Eno resources relative to the current time.
+// The idea is to identify deadlock states so they can be alerted on.
+type watchdogController struct {
+	client    client.Client
+	threshold time.Duration
+}
+
+func NewController(mgr ctrl.Manager, threshold time.Duration) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&apiv1.Composition{}).
```
AYM1607, line 22: The lifecycle controller takes care of this for Pods right? Or I guess that's implicit here, if Pods are stuck then the composition will also be stuck

internal/controllers/watchdog/controller.go
```diff
@@ -0,0 +1,71 @@
...
+		WithLogConstructor(manager.NewLogConstructor(mgr, "watchdogController")).
+		Complete(&watchdogController{
+			client:    mgr.GetClient(),
+			threshold: threshold,
+		})
+}
+
+func (c *watchdogController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	list := &apiv1.CompositionList{}
+	err := c.client.List(ctx, list)
```
AYM1607, line 32: I assume that it's safe to list them all because this is coming from the informer, but just want to sanity check.

internal/controllers/watchdog/controller.go
jveski: Exactly. Ideally the lifecycle controllers gets pods unstuck, but if not then it will be reflected here

AYM1607 (approved):
LGTM

PR #96, opened 2024-03-28
Limit the scope of configured selectors

jveski:
Currently we allow users to set default label/field selectors, but that can exposes a pretty significant space of impossible configurations, since one could set a label selector that matches a composition but not its resource slices, synthesis pods, etc.

So let's scope down those configurations: only allow setting a namespace and label selector _on compositions_ from only the reconciler. The controller can't really support selectors easily, and doesn't benefit much from them either.

### DIFF
--- a/cmd/eno-controller/main.go
+++ b/cmd/eno-controller/main.go
@@ -9,7 +9,6 @@ import (
 	"github.com/go-logr/zapr"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-	"k8s.io/apimachinery/pkg/labels"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/Azure/eno/internal/controllers/synthesis"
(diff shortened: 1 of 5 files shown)


cmd/eno-controller/main.go
```diff
@@ -39,13 +42,26 @@ func run() error {
 	flag.StringVar(&synconf.PodServiceAccount, "synthesizer-pod-service-account", "", "Service account name to be assigned to synthesizer Pods.")
 	flag.BoolVar(&debugLogging, "debug", true, "Enable debug logging")
 	flag.DurationVar(&watchdogThres, "watchdog-threshold", time.Minute*5, "How long before the watchdog considers a mid-transition resource to be stuck")
+	flag.StringVar(&compositionSelector, "composition-label-selector", "", "Optional label selector for compositions to be reconciled")
+	flag.StringVar(&compositionNamespace, "composition-namespace", "", "Optional namespace to limit compositions that will be reconciled")
```
AYM1607, line 46: default this to metav1.NamespaceAll to make the intent explicit?

cmd/eno-controller/main.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/go-logr/zapr"
```
AYM1607, line 9: Shouldn't the changes in this file be in the reconciler instead?

cmd/eno-controller/main.go
jveski: I like it - done

cmd/eno-controller/main.go
jveski: lol 😭

AYM1607 (approved):
LGTM

PR #97, opened 2024-03-29
Refactor reconstitution package

jveski:
This one got a bit gnarly, but I tried to keep the commit history coherent enough to track.

The highlights:

- Write buffer is decoupled and moved to separate package
- Remove support for multiple reconstitution queues since we ended up using one loop for reconciliation and status
- Move queue processor into its own file and address TODO around retry backoffs


cmd/eno-reconciler/main.go
```diff
@@ -106,10 +102,20 @@ func run() error {
 			remoteConfig.QPS = float32(remoteQPS)
 		}
 	}
-	err = reconciliation.New(recmgr, remoteConfig, discoveryMaxRPS, rediscoverWhenNotFound, readinessPollInterval)
+
+	writeBuffer := flowcontrol.NewResourceSliceWriteBufferForManager(mgr, writeBatchInterval, 1)
```
AYM1607, line 106: Burst went down from 2 to 1? Should it be adjustable?

jveski: I think 1 makes more sense - send 1 request immediately when something changes, then wait the debounce interval. So in cases where only a single resource changes we get quick feedback.

Personally I think we're okay hardcoding this for the sake of minimizing flags, but open to adding a flag if you feel strongly.

AYM1607: I think it's okay to hardcode it, I was just curious about the change, may be worth adding a comment with the fillowing?
> send 1 request immediately when something changes, then wait the debounce interval. So in cases where only a single resource changes we get quick feedback.

jveski: Pushed!

cmd/eno-reconciler/main.go
```diff
@@ -109,10 +104,18 @@ func run() error {
 	}
 
 	writeBuffer := flowcontrol.NewResourceSliceWriteBufferForManager(mgr, writeBatchInterval, 1)
-	err = reconciliation.New(recmgr, writeBuffer, remoteConfig, discoveryMaxRPS, rediscoverWhenNotFound, readinessPollInterval)
+
+	var reconciler reconciliation.Controller
```
AYM1607, line 110: I don't know if I love this pointer dance 😄 but I also can't think off the top of my head of a better way of doing it, I'll give it some thought though.

jveski: Good call! Fixed

AYM1607 (approved):
LGTM

PR #99, opened 2024-04-01
Add Symphony CRD

jveski:
Symphony allows sets of compositions to easily be spawned given a list of synthesizers. Kind of like a "meta composition" - a composition of compositions.


internal/controllers/replication/symphony.go
```diff
@@ -0,0 +1,223 @@
...
+	})
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("listing existing compositions: %w", err)
+	}
+
+	modified, err := c.syncStatus(ctx, symph, existing)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if modified {
```
AYM1607, line 54: What's the reasoning behind using modified as a signal to stop reconciliation of these sub-operations? Is it worth adding comments?

internal/controllers/replication/symphony.go
```diff
@@ -0,0 +1,223 @@
...
+	existingBySynthName, modified, err := c.reconcileReverse(ctx, symph, existing)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if modified {
+		return ctrl.Result{}, nil
+	}
+
+	if symph.DeletionTimestamp == nil {
+		modified, err := c.reconcileForward(ctx, symph, existingBySynthName)
```
AYM1607, line 76: Comments on why we need to reconcile forwards and backwards?

internal/controllers/replication/symphony.go
jveski: Passing down a bool is kind of strange looking, but it's the same behavior as other controllers: one write per tick of the loop.

This pattern guarantees that intermediate states will converge i.e. if the controller crashes between two writes, the behavior of the second write will be identical to if it hadn't crashed.

jveski: Added a comment

AYM1607 (approved):
LGTM

PR #103, opened 2024-04-08
More symphony-related improvements

jveski:
- Terminating pods shouldn't block composition deletion
- Symphony status fields should be nil if _any_ of the compositions have not reached that state. Currently it takes the max value.
- Don't re-delete resources in a couple of obscure cases
- Retry stale cache errors

That last one is tricky. When reading compositions from the resource slice status controller we error on 404s within 1min of resource slice creation. This is an excessively long time for informers to be sync'd, typically another event is received within a couple of ms. But there is one case in which this code can be reached legitimately: compositions that were deleted during synthesis. A resource slice may have already been written before deletion, so it will (rightly) not hold the composition finalizer. It's fine to defer GC'ing that slice for 1min but currently we never retry, so it will be blocked until resync.

### DIFF
--- a/internal/controllers/aggregation/symphony.go
+++ b/internal/controllers/aggregation/symphony.go
@@ -62,12 +62,11 @@ func (c *symphonyController) buildStatus(symph *apiv1.Symphony, comps *apiv1.Com
 	newStatus := apiv1.SymphonyStatus{ObservedGeneration: symph.Generation, Synthesizers: symph.Status.Synthesizers}
 
 	synthMap := map[string]struct{}{}
+	// Find the max values
 	for _, comp := range comps.Items {
-		synthMap[comp.Spec.Synthesizer.Name] = struct{}{}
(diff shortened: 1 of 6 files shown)


internal/controllers/aggregation/symphony.go
```diff
@@ -79,6 +78,27 @@ func (c *symphonyController) buildStatus(symph *apiv1.Symphony, comps *apiv1.Com
 		}
 	}
 
+	// Filter any values where one or more composition hasn't reached the corresponding state
```
AYM1607, line 81: So we're still taking the max, but we're just not reflecting it on the status until all compositions actually reach said status, right?

AYM1607 (approved):
LGTM

internal/controllers/aggregation/symphony.go
jveski: Yep exactly 👍

PR #108, opened 2024-04-19
Redesign rollout controller

jveski:
Moves the rollout controller into its own package and reworks it to enable rolling out other changes, like input changes.

This involves (vaguely):

- Removing rollout state from the synthesizer resource (doesn't make sense to track other rollouts on the synth)
- Using synthesis UUIDs to invalidate the reconstitution cache instead of the generation (the above item means generation won't increment during synth rollouts)

Also fixes a rollout bug afforded by a test not behaving quite like it reads.


cmd/eno-controller/main.go
```diff
@@ -41,6 +43,7 @@ func run() error {
 	flag.StringVar(&synconf.PodServiceAccount, "synthesizer-pod-service-account", "", "Service account name to be assigned to synthesizer Pods.")
 	flag.BoolVar(&debugLogging, "debug", true, "Enable debug logging")
 	flag.DurationVar(&watchdogThres, "watchdog-threshold", time.Minute*5, "How long before the watchdog considers a mid-transition resource to be stuck")
+	flag.DurationVar(&rolloutCooldown, "rollout-cooldown", time.Second*2, "How long before an update to related resource (synthesizer, bindings, etc.) will trigger a composition's re-synthesis")
```
AYM1607, line 46: This does not affect deferred inputs right?

AYM1607 (approved):
LGTM!

PR #111, opened 2024-04-25
Move discovery invalidation logic into reconciler

jveski:
Currently the reconciler process needs to be configured to handle schema cache invalidation correctly on different versions of apiserver. This change makes it so that clients don't need to know about this detail by moving the logic into the reconciler's discovery cache.

Also adds a 24hr cache invalidation to ensure that we eventually converge in cases where very old apiservers are upgraded.


AYM1607 (approved):
LGTM, nice!

PR #112, opened 2024-04-26
Add ✨ magical ✨ patch resource

jveski:
This new pseudo-resource allows resources not managed by Eno to be patched or deleted.

Also removes reconcileInterval from the resource slice and instead reconstitutes it at runtime from the resource manifest, like we do with all other values. Not a huge deal, but worth refactoring for the sake of consistency.


api/v1/synthesizer.go
```diff
@@ -16,6 +16,28 @@ type SynthesizerList struct {
 // Synthesizer processes are given some metadata about the composition they are synthesizing, and are expected
 // to return a set of Kubernetes resources. Essentially they generate the desired state for a set of Kubernetes resources.
 //
+// A special resource can be returned from synthesizers: `eno.azure.io/v1.Patch`.
```
AYM1607, line 19: We might need to add typing for this because for the synthesizer to return a correctly serialized Patch as unstructured, we need it in the schema

internal/resource/resource.go
```diff
@@ -87,6 +170,12 @@ func NewResource(ctx context.Context, renv *readiness.Env, slice *apiv1.Resource
 	return res, nil
 }
 
+type patchMeta struct {
+	APIVersion string          `json:"apiVersion"`
+	Kind       string          `json:"kind"`
+	Ops        jsonpatch.Patch `json:"ops"`
```
AYM1607, line 176: Slice of patches?

jveski: jsonpatch.Patch is (surprisingly) already a slice of patch operations.

api/v1/synthesizer.go
jveski: Hmm yeah it would be helpful to expose a scheme that has this type. Mind if that's a follow up? I have a couple of bug fixes that are going to be easier to land if we merge this first 😢

internal/resource/resource.go
```diff
@@ -39,6 +51,50 @@ func (r *Resource) Parse() (*unstructured.Unstructured, error) {
...
+	if err != nil {
+		return false
+	}
+
+	patchedjson, err := r.Patch.Apply(curjson)
+	if err != nil {
+		return false
+	}
+
+	patched := &unstructured.Unstructured{}
```
AYM1607, line 69: How about renaming this function to `GetPatched` and return both a bool and the patched resource to avoid recomputing it?

api/v1/synthesizer.go
AYM1607: No problem with a follow up :)

internal/resource/resource.go
jveski: IIRC (I wrote this on a Friday lol) this is the only place we apply the patch locally. The second application is when we send the raw patch json to apiserver.

internal/resource/resource.go
```diff
@@ -39,6 +51,50 @@ func (r *Resource) Parse() (*unstructured.Unstructured, error) {
...
+
+	return !equality.Semantic.DeepEqual(current, patched)
+}
+
+func (r *Resource) PatchDeletes() bool {
+	if r.Patch == nil {
+		return false
+	}
+
+	patchedjson, err := r.Patch.Apply([]byte(`{"apiVersion": "anything/v1", "kind":"Anything", "metadata":{}}`))
```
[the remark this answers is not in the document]
jveski: Just to satisfy the unstructured json parser. We apply the patch to an essentially empty resource and then check if the deletion timestamp is set as a way of determining if the patch signals deletion of the resource.

Added some comments and renamed the func for clarity.

internal/controllers/reconciliation/controller.go
```diff
@@ -177,7 +176,7 @@ func (c *Controller) reconcileResource(ctx context.Context, comp *apiv1.Composit
 		reconciliationLatency.Observe(time.Since(start).Seconds())
 	}()
 
-	if resource.Deleted() {
+	if resource.Deleted() || (resource.Patch != nil && resource.PatchSetsDeletionTimestamp()) {
```
AYM1607, line 179: Does it make sense to move the rest of this condition to `Deleted`? Or move it and rename to `ShouldDelete`?

jveski: Nice - fixed 👍

AYM1607 (approved):
LGTM

PR #113, opened 2024-04-29
More misc. bug fixes

jveski:
The main symptom these fixes address is stuck synth update rollouts. There are essentially three bugs that contribute to this, addressed in separate commits.

This also includes a change to include the API group in resource references, which in turn adds it to their index key. Resources are unique per kind+group, so this change solves for a rare case where the same type is used by resources in multiple groups.

### DIFF
--- a/internal/controllers/rollout/controller.go
+++ b/internal/controllers/rollout/controller.go
@@ -88,7 +88,7 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 
 		logger.V(1).Info("advancing rollout process")
-		return ctrl.Result{}, nil
+		return ctrl.Result{Requeue: true}, nil
 	}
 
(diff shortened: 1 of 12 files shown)


internal/controllers/rollout/controller.go
```diff
@@ -58,12 +58,6 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	var latestRollout time.Time
 	for _, comp := range compList.Items {
-		// For every synthesizer, only one composition can be the target of a rolling change at any point in time.
-		// To avoid blocked rollouts caused by compositions that cannot be synthesized, also time out and move on eventually.
-		if isRolling(&comp) {
```
jveski, line 63: This was always just a perf optimization but removing it isn't a huge deal. Better than risk more race conditions 😬

AYM1607 (approved):
LGTM

PR #114, opened 2024-04-30
List and watch by namespace

AYM1607:
Make sure that index and list operations respect namespaced resources

### DIFF
--- a/internal/manager/indices.go
+++ b/internal/manager/indices.go
@@ -1,0 +1,66 @@
+package manager
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	IdxPodsByComposition           = ".podsByComposition"
+	IdxCompositionsBySynthesizer   = ".spec.synthesizer"
+	IdxCompositionsBySymphony      = ".compositionsBySymphony"
+	IdxResourceSlicesByComposition = ".resourceSlicesByComposition"
+
+	CompositionNameLabelKey      = "eno.azure.io/composition-name"
+	CompositionNamespaceLabelKey = "eno.azure.io/composition-namespace"
+)
+
+func PodReferencesComposition(pod *corev1.Pod) bool {
+	labels := pod.GetLabels()
+	if labels == nil || labels[CompositionNameLabelKey] == "" || labels[CompositionNamespaceLabelKey] == "" {
+		return false
+	}
+	return true
+}
+
+func PodToCompMapFunc(ctx context.Context, obj client.Object) []reconcile.Request {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		logr.FromContextOrDiscard(ctx).V(0).Info("unexpected type given to podToCompMapFunc")
+		return nil
+	}
+	if !PodReferencesComposition(pod) {
+		return nil
+	}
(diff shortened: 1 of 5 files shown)


internal/manager/manager.go
```diff
@@ -117,12 +108,18 @@ func newMgr(logger logr.Logger, opts *Options, isController, isReconciler bool)
...
 
 	if isReconciler {
-		mgrOpts.Cache.ByObject[&apiv1.Composition{}] = cache.ByObject{
-			Namespaces: map[string]cache.Config{
-				opts.CompositionNamespace: {
-					LabelSelector: opts.CompositionSelector,
+		if opts.CompositionNamespace != cache.AllNamespaces {
+			mgrOpts.Cache.ByObject[&apiv1.Composition{}] = cache.ByObject{
+				Namespaces: map[string]cache.Config{
+					opts.CompositionNamespace: {
```
AYM1607, line 114: For some reason, setting this value to `AllNamespaces` causes "cache not configured for this namespace" errors, if I understand the implementation correctly, it shouldn't. This works for now, but suggestions are welcome, it should only be a problem in testing code anyways...

internal/manager/indices.go
```diff
@@ -0,0 +1,66 @@
...
+	return []reconcile.Request{{
+		NamespacedName: types.NamespacedName{
+			Name:      pod.GetLabels()[CompositionNameLabelKey],
+			Namespace: pod.GetLabels()[CompositionNamespaceLabelKey],
+		},
+	}}
+}
+
+func PodByCompIdxValueFromPod(po client.Object) string {
+	labels := po.GetLabels()
```
jveski, line 50: Handle nil? A few other places too.

(sidenote: it sure would be nice if these getters would return an empty map on nil)

internal/manager/indices.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	IdxPodsByComposition           = ".podsByComposition"
```
jveski, line 15: It's not important to include in this PR, but I'm thinking it would be nice to make these private and export funcs that return the client.MatchingFields for a given index while taking a structured input like a composition, etc. Kind of keeps the dynamic typing of the indices confined to this package.

internal/controllers/reconciliation/symphony_test.go
```diff
@@ -74,9 +78,40 @@ func TestSymphonyIntegration(t *testing.T) {
 	symph.Spec.Variations = []apiv1.Variation{{Synthesizer: apiv1.SynthesizerRef{Name: syn.Name}}}
 	require.NoError(t, upstream.Create(ctx, symph))
 
+	symph2 := &apiv1.Symphony{}
```
jveski, line 81: nit: would be nice to have a comment explaining what this resource is intended to cover. Or maybe a dedicated test case for testing namespace isolation.

internal/manager/manager.go
jveski: Yeah I'm not sure that's expected behavior from controller-runtime. In any case though this workaround isn't bad

(feel free to defer this and just add a TODO) We should add a test case to make sure the workaround doesn't regress in case the symphony test changes to no longer cover this functionality.

internal/controllers/reconciliation/symphony_test.go
AYM1607: Added a comment and a TODO to move this to a separate scenario.

internal/manager/manager.go
AYM1607: Added TODO.

jveski (approved):

PR #117, opened 2024-04-30
Misc. Flake fixes

AYM1607:
- Tolerate multiple syntheses in the controller tests, this only happens in testing code because Pods are (too?) fast.
- Observe the resource-version for a given resource only after successful reconciliation.


jveski (approved):
🏃

PR #118, opened 2024-04-30
Fix race between synthesis status and exec controllers

jveski:
It's possible for exec to complete before the status controller has written the synth generation to the status. This results in really weird behavior since nothing expects generation=0 to be possible.


AYM1607 (approved):
🏃

PR #120, opened 2024-05-02
Add callback from the write buffer

jveski:
To implement the resource ordering feature we will need to requeue _after_ status updates. This change just adds the hook


internal/flowcontrol/writebuffer.go
```diff
@@ -21,9 +21,10 @@ import (
 
 type StatusPatchFn func(*apiv1.ResourceState) *apiv1.ResourceState
 
-type ResourceSliceStatusUpdate struct {
+type resourceSliceStatusUpdate struct {
 	SlicedResource *reconstitution.ManifestRef
```
AYM1607, line 25: Do the fields need to be kept public now that this is unexported?

AYM1607 (approved):

PR #121, opened 2024-05-02
Parse readiness group annotation

jveski:
Adds `eno.azure.io/readiness-group`, which will be used to order the reconciliation of some resources after the readiness of others.

Resources without the annotation are expected to become ready first, then any resources in readiness-group=1 can be reconciled, continuing from there. Gaps in readiness group IDs are tolerated.

### DIFF
--- a/internal/resource/resource.go
+++ b/internal/resource/resource.go
@@ -44,7 +44,7 @@ type Resource struct {
 	ReadinessChecks   readiness.Checks
 	Patch             jsonpatch.Patch
 	DisableUpdates    bool
-	ReadinessGroup    uint8
+	ReadinessGroup    uint
 }
 
 func (r *Resource) Deleted() bool {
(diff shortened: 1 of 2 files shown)


internal/resource/resource.go
```diff
@@ -155,7 +157,12 @@ func NewResource(ctx context.Context, renv *readiness.Env, slice *apiv1.Resource
 	res.DisableUpdates = anno[disableUpdatesKey] == "true"
 	delete(anno, disableUpdatesKey)
 
-	for key, value := range parsed.GetAnnotations() {
+	const readinessGroupKey = "eno.azure.io/readiness-group"
+	rg, _ := strconv.ParseUint(anno[readinessGroupKey], 10, 0)
```
AYM1607, line 161: Why not bitSize 8 and handle the error? Also, why make the readiness groups so limited? Performance?

jveski: My initial implementation just bruteforced the readiness groups, so keeping the space small was significant. But I have since swapped that out for a red/black tree so we might as well use a "full sized" uint.

Also added code to log the errors, what do you think about continuing vs failing when things fail to parse? My concern is that we don't have any notion of terminal synth errors that cannot be fixed with retries, so we'd end up stuck in a synthesis loop.

AYM1607: Without a notion or terminal errors, I think continuing and just defaulting to 0 is the best we can do.

AYM1607 (approved):
lgtm

PR #122, opened 2024-05-02
Move manifest ref from Request to Resource

jveski:
Reconciliation queue work items should be deduplicated by the resource's group/kind/namespace/name (GKNN?) but currently they are also unique to a given resource slice slot.

So effectively we allow a workqueue item to be enqueued for both the previous and current version of the same resource, which is unnecessary and redundant.


AYM1607 (approved):
🥇

PR #124, opened 2024-05-03
Index resources by readiness group

jveski:
Uses a red-black tree to index resources by their readiness group. Lookups are cheap, so the reconciliation controller doesn't need tricks like maintaining its own representation of the rollout state.


AYM1607 (approved):
Nice!

PR #127, opened 2024-05-07
Support "readiness groups" i.e. dependencies

jveski:
Implements the concept of readiness groups.

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -105,9 +105,12 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 
 	var prev *reconstitution.Resource
 	if comp.Status.PreviousSynthesis != nil {
-		synRef.UUID = comp.Status.PreviousSynthesis.UUID
-		prev, _ = c.resourceClient.Get(ctx, synRef, &req.Resource)
+		prevSynRef := reconstitution.NewSynthesisRef(comp)
+		prevSynRef.UUID = comp.Status.PreviousSynthesis.UUID
+		prev, _ = c.resourceClient.Get(ctx, prevSynRef, &req.Resource)
 	}
+	logger = logger.WithValues("resourceKind", resource.Ref.Kind, "resourceName", resource.Ref.Name, "resourceNamespace", resource.Ref.Namespace)
+	ctx = logr.NewContext(ctx, logger)
 
 	// Keep track of the last reconciliation time and report on it relative to the resource's reconcile interval
 	// This is useful for identifying cases where the loop can't keep up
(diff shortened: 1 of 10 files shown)


internal/controllers/reconciliation/controller.go
```diff
@@ -105,9 +105,12 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 
 	var prev *reconstitution.Resource
 	if comp.Status.PreviousSynthesis != nil {
-		synRef.UUID = comp.Status.PreviousSynthesis.UUID
-		prev, _ = c.resourceClient.Get(ctx, synRef, &req.Resource)
+		prevSynRef := reconstitution.NewSynthesisRef(comp)
```
jveski, line 108: Avoid mutating the synthesis ref struct by just constructing a new one first here.

internal/reconstitution/queueprocessor.go
```diff
@@ -30,7 +30,7 @@ func (q *queueProcessor) processQueueItem(ctx context.Context) bool {
 	}
 	defer q.Queue.Done(item)
 
-	req, ok := item.(*Request)
+	req, ok := item.(Request)
```
jveski, line 33: Kind of unrelated change: using pointers breaks the workqueues de-duplication since it relies on equality checks.

internal/reconstitution/controller.go
```diff
@@ -105,3 +117,45 @@ func (r *controller) populateCache(ctx context.Context, comp *apiv1.Composition,
...
+			SliceName: slice.Name,
+			Namespace: slice.Namespace,
+		})
+		if !ok {
+			logger.V(1).Info("a dependent resource was not found in cache - this is unexpected")
+			return ctrl.Result{}, nil
+		}
+
+		synRef := &SynthesisRef{CompositionName: owner.Name, Namespace: req.Namespace, UUID: slice.Spec.SynthesisUUID}
+		resources := r.Cache.RangeByReadinessGroup(ctx, synRef, res.ReadinessGroup, 1)
```
AYM1607, line 150: Const (and maybe type?) for the direction? It's a bit hard to read at first glance

internal/controllers/reconciliation/controller.go
```diff
@@ -146,17 +138,49 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
...
+	if (status == nil || !status.Reconciled) && !resource.Deleted() {
+		dependencies := c.resourceClient.RangeByReadinessGroup(ctx, synRef, resource.ReadinessGroup, -1)
+		for _, dep := range dependencies {
+			slice := &apiv1.ResourceSlice{}
+			err = c.client.Get(ctx, dep.ManifestRef.Slice, slice)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("getting resource slice: %w", err)
+			}
+			status := dep.FindStatus(slice)
+			if status == nil || !status.Reconciled {
```
AYM1607, line 161: Reconciled -> Ready?

internal/controllers/reconciliation/controller.go
```diff
@@ -146,17 +138,49 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
...
 		if ok {
 			ready = &readiness.ReadyTime
 		}
+	} else {
+		ready = status.Ready
+	}
+
+	// Evaluate the readiness of resources in the previous readiness group
+	if (status == nil || !status.Reconciled) && !resource.Deleted() {
+		dependencies := c.resourceClient.RangeByReadinessGroup(ctx, synRef, resource.ReadinessGroup, -1)
```
AYM1607, line 153: Correct me if I'm wrong, but the way I'm reading this and the indexing PR, these are just dependencies within the same composition; will we also have readiness groups across compositions?

AYM1607: I totally see the use of intra-composition dependencies (get etcd ready before api-server in control-plane-core for ex) but we should also run through the possible scenarios where a composition might depend on another one

jveski: This is an interesting thought - let's discuss elsewhere

internal/controllers/reconciliation/controller.go
jveski: Good catch!

AYM1607 (approved):
LGTM!

internal/controllers/reconciliation/controller.go
AYM1607: Discussed offline, will stick with intra-composition readiness for now, we'll evaluate if inter-composition is required in the future

PR #128, opened 2024-05-08
Fix deletion when synth is missing

jveski:
Currently `kubectl delete`ing the examples never converges because the synth will disappear before the finalizer is removed from the composition, and the lifecycle controller bails out if the synth doesn't exist.


internal/controllers/synthesis/lifecycle.go
```diff
@@ -76,13 +77,23 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
 	}
 
+	// Tolerate missing synths since we may still need to cleanup
 	syn := &apiv1.Synthesizer{}
 	syn.Name = comp.Spec.Synthesizer.Name
 	err = c.client.Get(ctx, client.ObjectKeyFromObject(syn), syn)
+	// It's only safe to ignore as a missing synth if we have already started synthesis,
+	// otherwise creating the synth and composition around the same time could result in a deadlock
+	// if the composition is processed before the synth hits the informer cache.
+	if (errors.IsNotFound(err) || syn.DeletionTimestamp != nil) && comp.Status.CurrentSynthesis != nil {
```
AYM1607, line 87: Just out of curiosity, why check both the deletion timestamp and for a NotFound error? Could proceeding with a synth marked for deletion lead to a deadlock?

AYM1607 (approved):

internal/controllers/synthesis/lifecycle.go
jveski: This is maybe excessively defensive 😆 . I'm checking deletion timestamp here in case another component added a finalizer - the idea is to avoid trying to synthesize with a synthesizer that is actively being deleted.

AYM1607: Is there such a thing as excessively defensive? 👀  Looks good on my end and makes sense

PR #134, opened 2024-06-03
Pod spec improvements

jveski:
Makes some vaguely security related improvements and adds support for forcing synthesizer pods onto tainted nodes.

### DIFF
--- a/internal/controllers/synthesis/pod.go
+++ b/internal/controllers/synthesis/pod.go
@@ -4,6 +4,8 @@ import (
 	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 
 	apiv1 "github.com/Azure/eno/api/v1"
 	"github.com/Azure/eno/internal/manager"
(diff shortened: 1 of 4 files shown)


internal/controllers/synthesis/pod.go
```diff
@@ -25,42 +27,74 @@ func newPod(cfg *Config, comp *apiv1.Composition, syn *apiv1.Synthesizer) *corev
...
+			},
+		},
 		Containers: []corev1.Container{{
 			Name:    "synthesizer",
 			Image:   syn.Spec.Image,
 			Command: []string{"sleep", "infinity"},
 			SecurityContext: &corev1.SecurityContext{
+				AllowPrivilegeEscalation: ptr.To(false),
+				ReadOnlyRootFilesystem:   ptr.To(true),
+				RunAsUser:                ptr.To(int64(65532)),
```
AYM1607, line 54: This is not guaranteed to be correct, but I _think_ that it's not a big deal: worst case scenario, the Pod doesn't use distroless or equivalent as a base image which _hopefully_ leads to a terminal error. In any case it's only a problem for OSS 😜

internal/controllers/synthesis/pod.go
```diff
@@ -25,42 +27,74 @@ func newPod(cfg *Config, comp *apiv1.Composition, syn *apiv1.Synthesizer) *corev
...
+				RunAsGroup:               ptr.To(int64(65532)),
+				RunAsNonRoot:             ptr.To(true),
 				Capabilities: &corev1.Capabilities{
 					Drop: []corev1.Capability{"ALL"},
 				},
-				RunAsUser:    &userID,
-				RunAsNonRoot: &yes,
-			},
-			Env: []corev1.EnvVar{
```
AYM1607, line 43: Do you think there's any scenario where this information would be useful within the synth? Either way I think passing this through KRM is a better way of making this available.

internal/controllers/synthesis/pod_test.go
```diff
@@ -0,0 +1,90 @@
...
+	},
+	{
+		Name: "with affinity key/value",
+		Cfg: &Config{
+			NodeAffinityKey:   "foo",
+			NodeAffinityValue: "bar",
+		},
+		Assert: func(t *testing.T, p *corev1.Pod) {
+			require.NotNil(t, p.Spec.Affinity.NodeAffinity)
+			assert.Equal(t, corev1.NodeSelectorOpIn, p.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0].MatchExpressions[0].Operator)
```
AYM1607, line 37: assert on the value?

internal/controllers/synthesis/pod_test.go
```diff
@@ -0,0 +1,90 @@
...
+	},
+	{
+		Name: "with toleration key/value",
+		Cfg: &Config{
+			TaintTolerationKey:   "foo",
+			TaintTolerationValue: "bar",
+		},
+		Assert: func(t *testing.T, p *corev1.Pod) {
+			require.Len(t, p.Spec.Tolerations, 1)
+			assert.Equal(t, corev1.TolerationOpEqual, p.Spec.Tolerations[0].Operator)
```
AYM1607, line 48: Ditto

internal/controllers/synthesis/pod.go
jveski: At least the name of the composition is probably useful - definitely prefer passing it through KRM though. Will make a TODO for this

jveski:
Talked offline with Mariano yesterday - he's okay with merging but out of office today so I'll just slam this one in myself 🚀

PR #138, opened 2024-06-05
Reduce conflicts

jveski:
There are two controllers that often throw conflict errors since they update resources concurrent to other controllers. In one case there is no risk of meaningful conflict so we can just patch. In the other case (status controller) there is technically a condition in which the ordering matters so we'll have to retry conflicts. This could be refactored out in the future.


jveski:
This is pretty non-controversial - I'm going to merge while Mariano's out and we'll talk through the backlog of PRs when he's back. 🚢

PR #139, opened 2024-06-06
Switch from exec to process executor binary

jveski:
Exec'ing into the container to execute the synthesizer is non-ideal in a few ways: perf (latency) is not great, tracking retry state is complicated and historically racy, synthesizers have to retrieve their own inputs since we can't risk pushing too much through apiserver, etc. Shifting the same code into a wrapper binary in the container actually simplifies things pretty significantly and improves performance.


api/v1/inputs.go
```diff
@@ -78,6 +78,11 @@ type Ref struct {
 // A reference to a resource kind/group.
 type ResourceRef struct {
 	// +required
-	Kind  string `json:"kind"`
 	Group string `json:"group,omitempty"`
+
+	// +required
+	Version string `json:"version,omitempty"`
```
jveski, line 84: I changed my mind on this - I think we should specify a version instead of always returning latest

cmd/eno-controller/main.go
```diff
@@ -3,31 +3,47 @@ package main
...
 	"github.com/Azure/eno/internal/controllers/rollout"
 	"github.com/Azure/eno/internal/controllers/synthesis"
 	"github.com/Azure/eno/internal/controllers/watchdog"
+	"github.com/Azure/eno/internal/execution"
 	"github.com/Azure/eno/internal/manager"
 )
 
 func main() {
-	if err := run(); err != nil {
+	if len(os.Args) > 1 && os.Args[1] == "install-executor" {
```
jveski, line 31: The controller binary is reused for the "executor" process. Maybe we should release a separate image later, but it's so small I don't think it's worth the effort currently.

internal/controllers/reconciliation/status_test.go
```diff
@@ -177,67 +179,3 @@ func TestReconcileStatus(t *testing.T) {
 }
 
 func isNotReady(state apiv1.ResourceState) bool { return state.Ready == nil }
-
-// TestStatusNegative covers a weird corner case where a resource fails a patch but is still marked as reconciled.
```
jveski, line 181: This is no longer needed since the corner case it tested is just... not possible any more.

internal/controllers/synthesis/backoff_test.go
```diff
@@ -39,60 +37,10 @@ func TestControllerBackoff(t *testing.T) {
...
+			return comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Attempts >= 10
 		})
 
-		// It shouldn't be possible to try three times within 250ms
+		// It shouldn't be possible to try this many times within 250ms
 		assert.Greater(t, int(time.Since(start).Milliseconds()), 250)
 	})
 }
-
-func TestControllerBackoffReset(t *testing.T) {
```
jveski, line 50: This is no longer useful since the reset mechanism is very obvious, and integration testing it would be cumbersome.

jveski:

PR #140, opened 2024-06-07
Add simplified status in kubectl columns

jveski:
When interacting with compositions it's often difficult to tell which state they are in without squinting at the status properties for a while. This change adds a simple string representation of the composition's status and adds it as a kubectl column.


jveski:

PR #141, opened 2024-06-07
Support watching inputs

jveski:
Adds a "meta controller" that spawns a controller per unique `ref`, which resynthesizes compositions when their bound input resources change. Minimal implementation for now, will fill out more functionality (cooldown, etc.) later.


jveski:

PR #147, opened 2024-06-11
Support terminal synthesis errors

jveski:
Synthesizers can now return an error result to signal that the contents of the synthesis should be ignored.

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -87,7 +87,7 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 	}
 	logger := logr.FromContextOrDiscard(ctx).WithValues("compositionGeneration", comp.Generation)
 
-	if comp.Status.CurrentSynthesis == nil {
+	if comp.Status.CurrentSynthesis == nil || comp.Status.CurrentSynthesis.Failed() {
 		return ctrl.Result{}, nil // nothing to do
 	}
 	logger = logger.WithValues("synthesizerName", comp.Spec.Synthesizer.Name, "synthesizerGeneration", comp.Status.CurrentSynthesis.ObservedSynthesizerGeneration)
--- a/internal/controllers/rollout/controller.go
+++ b/internal/controllers/rollout/controller.go
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	apiv1 "github.com/Azure/eno/api/v1"
+	"github.com/Azure/eno/internal/controllers/synthesis"
 	"github.com/Azure/eno/internal/manager"
 )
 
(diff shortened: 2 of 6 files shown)


internal/controllers/rollout/controller.go
```diff
@@ -94,17 +95,6 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return ctrl.Result{}, nil
 }
 
-func SwapStates(comp *apiv1.Composition) {
```
jveski, line 97: I missed one implementation of swapStates when refactoring the other day - now we're down to one implementation in the synthesis controller.

internal/controllers/reconciliation/controller.go
```diff
@@ -87,7 +87,7 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 	}
 	logger := logr.FromContextOrDiscard(ctx).WithValues("compositionGeneration", comp.Generation)
 
-	if comp.Status.CurrentSynthesis == nil {
+	if comp.Status.CurrentSynthesis == nil || comp.Status.CurrentSynthesis.Failed() {
```
jveski, line 90: This is about 90% of the actual change being made in this PR 😆

internal/controllers/synthesis/lifecycle.go
```diff
@@ -318,10 +318,10 @@ func shouldDeletePod(logger logr.Logger, comp *apiv1.Composition, syn *apiv1.Syn
 	return logger, nil, false
 }
 
-func swapStates(comp *apiv1.Composition) {
-	// If the previous state has been synthesized but not the current, keep the previous to avoid orphaning deleted resources
-	if comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Synthesized != nil {
-		comp.Status.PreviousSynthesis = comp.Status.CurrentSynthesis
+func SwapStates(comp *apiv1.Composition) {
+	current := comp.Status.CurrentSynthesis
+	if current != nil && current.Synthesized != nil && !current.Failed() {
```
jveski, line 323: This prevents us from losing the last known good state - otherwise we may end up orphaning resources, changing properties we shouldn't (due to missing three way merge metadata), etc.

internal/controllers/reconciliation/controller.go
AYM1607: Okay so this just doesn't reconcile the contents of a failed synthesis.

internal/controllers/synthesis/lifecycle.go
AYM1607: Essentially this is causing the failed synthesis to "stay current" and be overwritten with an "empty" synthesis?

jveski: Yep exactly. Otherwise we might push out a non-empty synthesis and lose the three-way merge state it contains.

internal/controllers/reconciliation/controller.go
jveski: Right. We could add a third synthesis to the composition status to avoid blocking reconciliation, but I think terminal errors are rare enough this is fine

AYM1607 (approved):

PR #148, opened 2024-06-13
Finish implementing deferred input rollouts

jveski:
Replaces the existing rollout controller with a new implementation that implements slightly nicer semantics (see reference docs for details).

### DIFF
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -43,9 +43,9 @@ type CompositionSpec struct {
 
 type CompositionStatus struct {
 	Simplified         *SimplifiedStatus `json:"simplified,omitempty"`
-	PendingResynthesis *metav1.Time      `json:"pendingResynthesis,omitempty"`
 	CurrentSynthesis   *Synthesis        `json:"currentSynthesis,omitempty"`
 	PreviousSynthesis  *Synthesis        `json:"previousSynthesis,omitempty"`
+	PendingResynthesis *metav1.Time      `json:"pendingResynthesis,omitempty"`
 }
 
 type SimplifiedStatus struct {
@@ -92,6 +92,10 @@ type Synthesis struct {
 
 	// InputRevisions contains the versions of the input resources that were used for this synthesis.
 	InputRevisions []InputRevisions `json:"inputRevisions,omitempty"`
+
+	// Deferred is true when this synthesis was the result of a deferred rollout,
+	// caused by a synthesizer of (deferred) input change.
+	Deferred bool `json:"deferred,omitempty"`
 }
 
 type Result struct {
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -93,8 +93,8 @@ type Synthesis struct {
 	// InputRevisions contains the versions of the input resources that were used for this synthesis.
 	InputRevisions []InputRevisions `json:"inputRevisions,omitempty"`
 
-	// Deferred is true when this synthesis was the result of a deferred rollout,
-	// caused by a synthesizer of (deferred) input change.
+	// Deferred is true when this synthesis was caused by a change to either the synthesizer
+	// or an input with a ref that sets `Defer == true`.
 	Deferred bool `json:"deferred,omitempty"`
 }
 
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -42,9 +42,10 @@ type CompositionSpec struct {
 }
 
 type CompositionStatus struct {
-	Simplified        *SimplifiedStatus `json:"simplified,omitempty"`
-	CurrentSynthesis  *Synthesis        `json:"currentSynthesis,omitempty"`
-	PreviousSynthesis *Synthesis        `json:"previousSynthesis,omitempty"`
+	Simplified         *SimplifiedStatus `json:"simplified,omitempty"`
+	PendingResynthesis *metav1.Time      `json:"pendingResynthesis,omitempty"`
+	CurrentSynthesis   *Synthesis        `json:"currentSynthesis,omitempty"`
+	PreviousSynthesis  *Synthesis        `json:"previousSynthesis,omitempty"`
 }
 
 type SimplifiedStatus struct {
--- a/cmd/eno-controller/main.go
+++ b/cmd/eno-controller/main.go
@@ -97,7 +97,7 @@ func runController() error {
 		return fmt.Errorf("constructing manager: %w", err)
 	}
 
-	err = rollout.NewController(mgr, rolloutCooldown)
+	err = rollout.NewSynthesizerController(mgr, rolloutCooldown)
 	if err != nil {
 		return fmt.Errorf("constructing rollout controller: %w", err)
 	}
--- a/cmd/eno-controller/main.go
+++ b/cmd/eno-controller/main.go
@@ -97,7 +97,12 @@ func runController() error {
 		return fmt.Errorf("constructing manager: %w", err)
 	}
 
-	err = rollout.NewSynthesizerController(mgr, rolloutCooldown)
+	err = rollout.NewController(mgr, rolloutCooldown)
+	if err != nil {
+		return fmt.Errorf("constructing rollout controller: %w", err)
+	}
+
+	err = rollout.NewSynthesizerController(mgr)
 	if err != nil {
 		return fmt.Errorf("constructing rollout controller: %w", err)
 	}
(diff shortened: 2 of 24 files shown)


api/v1/composition.go
```diff
@@ -91,6 +92,10 @@ type Synthesis struct {
 
 	// InputRevisions contains the versions of the input resources that were used for this synthesis.
 	InputRevisions []InputRevisions `json:"inputRevisions,omitempty"`
+
+	// Deferred is true when this synthesis was the result of a deferred rollout,
+	// caused by a synthesizer of (deferred) input change.
```
AYM1607, line 97: Nit: This phrasing is a bit unclear

cmd/eno-controller/main.go
```diff
@@ -64,7 +64,7 @@ func runController() error {
 	flag.StringVar(&synconf.PodServiceAccount, "synthesizer-pod-service-account", "", "Service account name to be assigned to synthesizer Pods.")
 	flag.BoolVar(&debugLogging, "debug", true, "Enable debug logging")
 	flag.DurationVar(&watchdogThres, "watchdog-threshold", time.Minute*5, "How long before the watchdog considers a mid-transition resource to be stuck")
-	flag.DurationVar(&rolloutCooldown, "rollout-cooldown", time.Second*2, "How long before an update to related resource (synthesizer, bindings, etc.) will trigger a composition's re-synthesis")
+	flag.DurationVar(&rolloutCooldown, "rollout-cooldown", time.Minute, "How long before an update to a related resource (synthesizer, bindings, etc.) will trigger a second composition's re-synthesis")
```
AYM1607, line 67: Does the "second" here means to imply that it only affects compositions that have been synthesized at least once?

dev/build.sh
```diff
@@ -22,5 +22,5 @@ done
 wait
 
 # Deploy!
-cat "$(dirname "$0")/deploy.yaml" | envsubst | kubectl apply -f -
+cat "$(dirname "$0")/deploy.yaml" | envsubst | kubectl apply -f - -f ./api/v1/config/crd
```
AYM1607, line 25: I didn't know you could do this 👀

docs/reference.md
```diff
@@ -215,3 +215,13 @@ patch:
...
+
+
+# Rollouts
+
+Composition changes are resynthesized immediately.
+Changes to deferred input resources (`ref.defer == true`) and synthesizers are subject to the global cooldown period.
+
+- All effected compositions are marked as pending resynthesis immediately
+- A maximum of one composition pending resynthesis can begin resynthesis per cooldown period
+- The next pending composition can start after the cooldown period has expired AND all resynthesis has completed or been retried at least once
```
AYM1607, line 227: I'd assume we can also move on to the next one if there's a terminal error? I'll probably find this out when I read the code 😄

internal/controllers/rollout/controller.go
```diff
@@ -25,114 +22,81 @@ type controller struct {
...
-			latestRollout = comp.Status.CurrentSynthesis.Synthesized.Time
+	// Find the current cooldown period, wait for the next if needed
+	// Block on any active deferred syntheses with less than 2 attempts to avoid high concurrency
+	var lastDeferredSynthesisCompletion *metav1.Time
+	for _, comp := range comps.Items {
+		current := comp.Status.CurrentSynthesis
+		if current == nil || !current.Deferred {
+			continue
+		}
+		if current.Synthesized == nil && current.Attempts < 2 {
```
AYM1607, line 54: Are we relying solely on exponential backoff to limit concurrency if we have a bunch of compositons being retried?

internal/controllers/rollout/synthesizer.go
```diff
@@ -0,0 +1,127 @@
...
+	compList := &apiv1.CompositionList{}
+	err = c.client.List(ctx, compList, client.MatchingFields{
+		manager.IdxCompositionsBySynthesizer: syn.Name,
+	})
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("listing compositions: %w", err)
+	}
+
+	// randomize list to avoid always rolling out changes in the same order
+	rand.Shuffle(len(compList.Items), func(i, j int) { compList.Items[i] = compList.Items[j] })
```
AYM1607, line 58: This is a behavior change compared to the OM's hash buckets, but we've discussed before that it's not a good idea anyways 😄

internal/controllers/rollout/synthesizer.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+		UpdateFunc: func(ctx context.Context, ue event.UpdateEvent, rli workqueue.RateLimitingInterface) {
+			newComp, ok := ue.ObjectNew.(*apiv1.Composition)
+			if !ok {
+				logr.FromContextOrDiscard(ctx).V(0).Info("unexpected type given to newCompositionToSynthesizerHandler")
+				return
+			}
+
+			oldComp, ok := ue.ObjectOld.(*apiv1.Composition)
+			if !ok {
```
AYM1607, line 111: Can this ever happen?

internal/controllers/rollout/synthesizer.go
```diff
@@ -0,0 +1,127 @@
...
+	// randomize list to avoid always rolling out changes in the same order
+	rand.Shuffle(len(compList.Items), func(i, j int) { compList.Items[i] = compList.Items[j] })
+
+	for _, comp := range compList.Items {
+		comp := comp
+		logger := logger.WithValues("compositionName", comp.Name, "compositionNamespace", comp.Namespace, "compositionGeneration", comp.Generation)
+
+		// Compositions aren't eligible to receive an updated synthesizer when:
+		// - They haven't ever been synthesized (they'll use the latest inputs anyway)
+		// - They are currently being synthesized or deleted
```
AYM1607, line 66: The current synthesis could be using the previous generation of the synthesizer, would it not get stuck there if there are no composition changes until the next synthesizer release?

internal/controllers/rollout/synthesizer.go
```diff
@@ -0,0 +1,127 @@
...
+		// - They are already pending resynthesis
+		// - They are already in sync with the latest inputs
```
AYM1607, line 68: "latest synthesizer"?

cmd/eno-controller/main.go
jveski: Right- the first synthesis will always be non-deferred, essentially because the composition's generation will transition from zero to one.

internal/controllers/rollout/controller.go
```diff
@@ -25,114 +22,81 @@ type controller struct {
...
 		client:   mgr.GetClient(),
 		cooldown: cooldown,
 	}
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&apiv1.Synthesizer{}).
-		Watches(&apiv1.Composition{}, newCompositionHandler()).
-		// TODO: Filter some events?
-		WithLogConstructor(manager.NewLogConstructor(mgr, "synthesizerRolloutController")).
+		Named("rolloutController").
+		Watches(&apiv1.Composition{}, manager.SingleEventHandler()).
```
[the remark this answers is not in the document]
jveski: We use `manager.SingleEventHandler` so that `RequeueAfter` results will be honored across compositions. Otherwise there's a risk of dispatching another synthesis before the previously dispatched one hits the informer.

internal/controllers/rollout/controller.go
jveski: Yeah for now. I still think we may need to add pod concurrency limits at some point though

internal/controllers/rollout/synthesizer.go
jveski: Yeah the next level would be to let compositions opt in to receive deferred changes earlier 😸

internal/controllers/rollout/synthesizer.go
jveski: The previous resource isn't sent from the server in watch events, so (if memory serves 😬 ) it's populated with the last version in the informer cache when the new version is received. So it can be nil if the process has crashed.

internal/controllers/rollout/synthesizer.go
jveski: The synthesized time and/or uuid will change, which will cause this code to be executed again. So we'll always catch up eventually.

AYM1607 (approved):

PR #154, opened 2024-07-01
Misc. resilience improvements

jveski:
- Limits synthesis concurrency
- Releases resource slices from previous synthesis attempts to prevent accumulation during retry loops
- Replaces some full status updates with patches in cases where retrying conflicts adds meaningful overhead and other concurrency controls are in place

### DIFF
--- a/internal/controllers/flowcontrol/synthesis.go
+++ b/internal/controllers/flowcontrol/synthesis.go
@@ -1,0 +1,81 @@
+package flowcontrol
+
+import (
+	"context"
+	"fmt"
+
+	apiv1 "github.com/Azure/eno/api/v1"
+	"github.com/Azure/eno/internal/manager"
+	"github.com/go-logr/logr"
+	"github.com/google/uuid"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type synthesisConcurrencyLimiter struct {
+	client client.Client
+	limit  int
+}
+
+func NewSynthesisConcurrencyLimiter(mgr ctrl.Manager, limit int) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("synthesisConcurrencyLimiter").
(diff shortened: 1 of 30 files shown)


internal/controllers/flowcontrol/synthesis.go
```diff
@@ -0,0 +1,81 @@
...
+	err := c.client.List(ctx, list)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	var active int
+	var pending int
+	var next *apiv1.Composition
+	for _, comp := range list.Items {
+		comp := comp
```
AYM1607, line 44: Are we on go 1.22? If so this is no longer needed

internal/controllers/flowcontrol/synthesis.go
```diff
@@ -0,0 +1,81 @@
...
+		current := comp.Status.CurrentSynthesis
+		if current != nil && current.Synthesized != nil {
+			continue // already synthesized
+		}
+		if current == nil || current.UUID == "" {
+			pending++
+			next = &comp
```
AYM1607, line 51: Should we sort (or shuffle) to have better guarantees around choosing the next comp to be dispatched?

internal/controllers/flowcontrol/synthesis.go
jveski: Still on 1.21 I believe, good reminder to update :)

internal/controllers/flowcontrol/synthesis.go
jveski: There's a risk of the loop "outpacing" the informer, since it will be re-entered by any composition change and not just changes caused by the controller. Currently it will fail closed by retrying the same composition and getting a conflict error, shuffling would potentially result in a few extras getting dispatched before the informer catches up.

I do think this is worth addressing, but I'll do so in a separate PR to keep the scope of this one somewhat reasonable

AYM1607 (approved):

PR #157, opened 2024-07-09
Allow synthesizer pod annotations/labels/resources to be set in synthesizer

jveski:
Adds "pod overrides" to the synthesizer CRD.


internal/controllers/synthesis/pod.go
```diff
@@ -57,6 +65,7 @@ func newPod(cfg *Config, comp *apiv1.Composition, syn *apiv1.Synthesizer) *corev
 				Name:      "sharedfs",
 				MountPath: "/eno",
 			}},
+			Resources: syn.Spec.PodOverrides.Resources,
```
AYM1607, line 68: Is the zero value safe to use here? Should we make it an optional pointer and only set it when it's non nil?

jveski: I slightly prefer staying consistent with the pod API which doesn't use pointers for these values. Since the zero value is meaningful here I don't think there's a significant reason to go either way - but happy to change if you have a stronger opinion

AYM1607: Zero value works for me, you are right, that'd already be the value if we were o not set it explicitly.

AYM1607 (approved):

PR #158, opened 2024-07-11
Wait for inputs to exist

jveski:
Currently synthesis pods will be created even if the bound resources do not exist, which just results in a lot of crashlooping.


internal/controllers/synthesis/integration_test.go
```diff
@@ -56,35 +56,39 @@ func TestControllerHappyPath(t *testing.T) {
 	comp.Spec.Synthesizer.Name = syn.Name
 	require.NoError(t, cli.Create(ctx, comp))
 
-	t.Run("initial creation", func(t *testing.T) {
```
jveski, line 59: small refactor away from t.Run since it isn't used quite correctly here

internal/controllers/synthesis/lifecycle.go
```diff
@@ -138,24 +139,8 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
-	// Before it's safe to create a pod, we need to write _something_ back to the composition.
-	// This will fail if another write has already hit the resource i.e. synthesis completion.
-	// Otherwise it's possible for pod deletion event to land before the synthesis complete event.
-	// In that case a new pod would be created even though the synthesis just completed.
-	//
-	// The UUID is written by the flow control controller - this controller should just wait
-	// until it has "dispatched" this synthesis.
-	if comp.Status.CurrentSynthesis.UUID == "" {
+	// Bail if it isn't time to synthesize yet, or synthesis is already complete
+	if comp.Status.CurrentSynthesis == nil || comp.Status.CurrentSynthesis.UUID == "" || comp.Status.CurrentSynthesis.Synthesized != nil || comp.DeletionTimestamp != nil {
```
jveski, line 143: small refactor to remove some rather obtuse comments and flatten the several branches into one

internal/controllers/synthesis/lifecycle.go
```diff
@@ -335,10 +320,15 @@ func SwapStates(comp *apiv1.Composition) {
 }
 
 func shouldSwapStates(comp *apiv1.Composition) bool {
-	// synthesize when:
-	// - the spec has changed since last synthesis
-	// - synthesis is complete, but the uuid is missing
```
jveski, line 340: this was only needed during the migration to synthesis uuid

internal/controllers/watch/pruning.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"fmt"
+
+	apiv1 "github.com/Azure/eno/api/v1"
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type pruningController struct {
```
jveski, line 13: Now that we write the input revisions to the composition status (not the current synthesis) we need to clean them up eventually

internal/controllers/flowcontrol/synthesis.go
```diff
@@ -43,10 +43,10 @@ func (c *synthesisConcurrencyLimiter) Reconcile(ctx context.Context, req ctrl.Re
 	for _, comp := range list.Items {
 		comp := comp
 		current := comp.Status.CurrentSynthesis
-		if current != nil && current.Synthesized != nil {
-			continue // already synthesized
+		if current == nil || current.Synthesized != nil {
```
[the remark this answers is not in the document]
AYM1607: A flowchart would be immensely helpful!

internal/controllers/aggregation/composition_test.go
```diff
@@ -54,6 +55,20 @@ func TestCompositionSimplification(t *testing.T) {
...
+				Status: "MissingInputs",
+			},
+		},
+		{
+			Bindings: []apiv1.Binding{{Key: "foo"}},
+			Input:    apiv1.CompositionStatus{CurrentSynthesis: &apiv1.Synthesis{UUID: "uuid"}},
+			Expected: apiv1.SimplifiedStatus{
+				Status: "MissingInputs",
+			},
+		},
```
AYM1607, line 71: Test with a binding and the expected status?

internal/controllers/synthesis/lifecycle.go
```diff
@@ -357,3 +346,32 @@ func shouldUpdateDeletedCompositionStatus(comp *apiv1.Composition) bool {
...
+	}
+
+	refsByKey := map[string]apiv1.Ref{}
+	for _, ref := range synth.Spec.Refs {
+		ref := ref
+		refsByKey[ref.Key] = ref
+	}
+
+	var equal int
+	for _, a := range a {
```
AYM1607, line 363: Mega nit, but maybe name the arguments ar/br and the loop variables a/b? Shadowing sometimes makes his hard to read, but it's just my preference so feel free to ignore.

internal/controllers/watch/watch.go
```diff
@@ -27,6 +27,16 @@ func NewController(mgr ctrl.Manager) error {
 			client:         mgr.GetClient(),
 			refControllers: map[apiv1.ResourceRef]*KindWatchController{},
 		})
+	if err != nil {
+		return err
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
```
AYM1607, line 34: Is it okay to just return the pruning controller here/

internal/controllers/watch/pruning.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (c *pruningController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	comp := &apiv1.Composition{}
+	err := c.client.Get(ctx, req.NamespacedName, comp)
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	for i, ir := range comp.Status.InputRevisions {
+		if hasBindingKey(comp, ir.Key) {
```
AYM1607, line 25: Just to be clear, there's always one input revision per input and this only prunes when the composition stops binding the input right?

jveski: Exactly

internal/controllers/watch/watch.go
jveski: `Complete()` returns an error, which is the only return value here. Any concerns (I might be unclear on the question)

AYM1607: Ah, I though the controller itself was being returned, lgtm.

AYM1607 (approved):

PR #166, opened 2024-07-19
Update getting-started doc

chihshenghuang:
Add CRD installation description

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -2,6 +2,8 @@
 
 ## Installing Eno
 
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -2,6 +2,8 @@
 
 ## Installing Eno
 
+At the begining, install the Eno CRD resource defined at [here](https://github.com/Azure/eno/tree/main/api/v1/config/crd) to your cluster.
```
jveski, line 5: Great catch. I'm thinking we could simplify onboarding a bit by concatenating the deploy.yaml manifest with the generated CRDs into a release artifact like `manifest.yaml`. Then swap out the file we curl in the steps below for that one

jveski (approved):

PR #167, opened 2024-07-19
Fix dispatch race

jveski:
It's possible for the same synthesis to be dispatch multiple times i.e. have its uuid changed. This manifests in different deadlock states, all of which seem to eventually recover assuming reasonable synthesis timeouts. So no real impact IMO just annoying test flakes.

### DIFF
--- a/internal/execution/executor.go
+++ b/internal/execution/executor.go
@@ -27,6 +27,8 @@ type Executor struct {
 }
 
 func (e *Executor) Synthesize(ctx context.Context, env *Env) error {
+	logger := logr.FromContextOrDiscard(ctx)
+
 	comp := &apiv1.Composition{}
 	comp.Name = env.CompositionName
 	comp.Namespace = env.CompositionNamespace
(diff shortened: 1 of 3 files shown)


internal/execution/executor.go
```diff
@@ -217,7 +219,19 @@ func (e *Executor) updateComposition(ctx context.Context, env *Env, oldComp *api
 	})
 }
 
-func skipSynthesis(comp *apiv1.Composition, env *Env) bool {
+func skipSynthesis(comp *apiv1.Composition, env *Env) (string, bool) {
 	synthesis := comp.Status.CurrentSynthesis
-	return synthesis == nil || synthesis.Synthesized != nil || synthesis.UUID != env.SynthesisUUID || (synthesis.Attempts > 0 && synthesis.Attempts > env.SynthesisAttempt)
+	if synthesis == nil {
```
jveski, line 224: Tested a theory by adding better logging. Theory didn't end up being true, but thought the logging might be useful for the future

chihshenghuang (approved):

PR #168, opened 2024-07-19
Fix test flake

jveski:
Another patch related race condition that (very likely) only effects tests 🤷

### DIFF
--- a/internal/controllers/synthesis/lifecycle.go
+++ b/internal/controllers/synthesis/lifecycle.go
@@ -2,6 +2,7 @@ package synthesis
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/controllers/synthesis/lifecycle.go
```diff
@@ -177,11 +179,18 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 	sytheses.Inc()
 
 	// This metadata is optional - it's safe for the process to crash before reaching this point
-	copy := comp.DeepCopy()
-	copy.Status.CurrentSynthesis.Attempts++
-	copy.Status.CurrentSynthesis.PodCreation = &pod.CreationTimestamp
-	err = c.client.Status().Patch(ctx, copy, client.MergeFrom(comp))
+	patch := []map[string]any{
+		{"op": "test", "path": "/status/currentSynthesis/uuid", "value": comp.Status.CurrentSynthesis.UUID},
```
AYM1607, line 183: Essentially moving these checks to the apiserver instead of doing them locally?

jveski: Exactly. The problem is that the normal optimistic concurrency control approach is... too optimistic. i.e. it's very likely that the resource version has changed at this point, but it's also likely that the resource itself hasn't changed in ways that _semantically_ conflict with this patch. So the `test` op allows us to preserve safety in extreme cases without sacrificing efficiency in the happy path

AYM1607 (approved):

PR #171, opened 2024-07-22
Fix smoke test logic

jveski:
The smoke tests pass if any single composition becomes ready 🤷 .

This fixes it and adds better logging.

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -11,11 +11,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Go
-        uses: actions/setup-go@v5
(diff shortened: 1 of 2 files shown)


.github/workflows/smoke.yaml
```diff
@@ -11,11 +11,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Go
```
AYM1607, line 14: I assume this is because the build is happening inside docker?

AYM1607 (approved):

.github/workflows/smoke.yaml
jveski: Yep!

PR #172, opened 2024-07-22
Add test for Eno->Helm migration

jveski:
Adds a test case to make sure Helm releases can adopt resources created by Eno (of course hopefully no one ever uses this 🙃 )

### DIFF
--- a/internal/controllers/reconciliation/helm_test.go
+++ b/internal/controllers/reconciliation/helm_test.go
@@ -154,3 +154,98 @@ func TestHelmOwnershipTransfer(t *testing.T) {
 		"app.kubernetes.io/managed-by": "Helm",
 	}, obj.GetLabels())
 }
+
+// TestHelmOwnershipTransfer is similar to TestHelmOwnershipTransfer but covers the adoption of resources initially created by Eno.
+func TestHelmOwnershipTransferAfterCreation(t *testing.T) {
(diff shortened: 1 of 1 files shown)


AYM1607:
I **really** hope no one uses this lol

internal/controllers/reconciliation/helm_test.go
```diff
@@ -154,3 +154,98 @@ func TestHelmOwnershipTransfer(t *testing.T) {
...
+		if comp.Status.CurrentSynthesis == nil || comp.Status.CurrentSynthesis.ObservedCompositionGeneration != comp.Generation || comp.Status.CurrentSynthesis.Reconciled == nil {
+			return false
+		}
+
+		obj.SetName("test-obj")
+		obj.SetNamespace("default")
+		err = downstream.Get(ctx, client.ObjectKeyFromObject(obj), obj)
+		uid = string(obj.GetUID())
+		return err == nil
+	})
```
AYM1607, line 237: For the sake of completeness, should we add a step here to delete the comp with orphaning set?

jveski: We cover that in the other helm test case, so I'd rather avoid it here. These tests are just so dang long already

AYM1607 (approved):

PR #176, opened 2024-07-24
Allow annotations passthrough from symphony variations to compositions

AYM1607:
Resource orphaning is set through an annotation, the symphony API currently only allows to pass labels.


go.work.sum
```diff
@@ -5,20 +5,27 @@ cloud.google.com/go/compute/metadata v0.3.0/go.mod h1:zFmK7XCadkQkj6TtorcaGlCW1h
 dario.cat/mergo v1.0.0/go.mod h1:uNxQE+84aUszobStD9th8a29P2fMDhsBdgRYvZOxGmk=
 github.com/Masterminds/vcs v1.13.3/go.mod h1:TiE7xuEjl1N4j016moRd6vezp6e6Lz23gypeXfzXeW8=
 github.com/NYTimes/gziphandler v1.1.1/go.mod h1:n/CVRwUEOgIxrgPvAQhUUr9oeUtvrhMomdKFjzJNB0c=
+github.com/agnivade/levenshtein v1.0.1/go.mod h1:CURSv5d9Uaml+FovSIICkLbAUZ9S4RqaHDIsdSBg7lM=
```
AYM1607, line 8: I've never worked with go workspaces, no idea if these are expected or not, but seem harmless

jveski (approved):

go.work.sum
jveski: Yeah same. I think somehow those were missed in my last PR 🤷

PR #179, opened 2024-07-24
Add input revision checks

jveski:
Currently we parse the input revision annotation but don't actually block synthesis per its implied logic. This PR adds the implementation and clarifies the behavior in the docs.


api/v1/composition.go
```diff
@@ -160,3 +160,29 @@ func (c *Composition) InputsExist(syn *Synthesizer) bool {
...
+	var minRevision *int
+	for _, rev := range c.Status.InputRevisions {
+		if rev.Revision == nil {
+			continue
+		}
+		if minRevision == nil {
+			minRevision = rev.Revision
+			continue
+		}
+		if *rev.Revision > *minRevision {
```
AYM1607, line 174: *rev.Revision < *minRevision?

AYM1607, line 174: `*rev.Revision > *minRevision`?

docs/reference.md
```diff
@@ -168,13 +168,12 @@ spec:
 
 The composition will be resynthesized whenever `test-input`'s `resourceVersion` changes by default, subject to the global cooldown period.
 
-It's possible to defer resynthesis until some other inputs have been modified.
-This is useful for cases in which another process updates several input resources serially.
-
-Setting this annotation to a higher value than any other input of a given composition will not result in resynthesis until all other inputs have the same value.
+If several inputs are expected to transition in lockstep, use this annotation to override the resource version.
+Synthesis will only happen once all inputs satisfy `revision >= max(revisions)` where `revisions` is scoped to all resources bound to that composition.
```
AYM1607, line 172: Out of curiosity, why `revision >= max(revisions)` and not `revision == max(revisions)`?

jveski: Because I wrote it during a meeting (shame) 🙃.

Of course this doesn't make logical sense - revision can't be > the max of revisions

api/v1/composition.go
jveski: I think this is right as-is, but `minRevision` should be named `maxRevision` :blush:

go.work.sum
```diff
@@ -3,9 +3,12 @@ cloud.google.com/go/compute v1.25.1/go.mod h1:oopOIR53ly6viBYxaDhBfJwzUAxf1zE//u
 cloud.google.com/go/compute/metadata v0.2.3/go.mod h1:VAV5nSsACxMJvgaAuX6Pk2AawlZn8kiOGuCv6gTkwuA=
 cloud.google.com/go/compute/metadata v0.3.0/go.mod h1:zFmK7XCadkQkj6TtorcaGlCW1hT1fIilQDwofLpJ20k=
 dario.cat/mergo v1.0.0/go.mod h1:uNxQE+84aUszobStD9th8a29P2fMDhsBdgRYvZOxGmk=
+github.com/AdamKorcz/go-118-fuzz-build v0.0.0-20230306123547-8075edf89bb0/go.mod h1:OahwfttHWG6eJ0clwcfBAHoDI6X/LV/15hx/wlMZSrU=
```
AYM1607, line 6: this go work file is going a little wild 😄

internal/controllers/synthesis/lifecycle.go
```diff
@@ -337,7 +337,7 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
 	// - synthesis is not already pending
 	// - all bound input resources exist (or composition is being deleted)
```
AYM1607, line 338: Update comment to reflect that the revisions must match.

AYM1607 (approved):

PR #180, opened 2024-07-25
Fix composition deletion when a resource is missing namespace

jveski:
If a resource returned by a synthesizer doesn't have a namespace, getting its current state will fail, which prevents it from being marked as deleted, which blocks composition deletion.

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


AYM1607:
Leaving this comment here to reflect an offline conversation: Currently Eno opaquely passes empty namespaces generated by synthesizers and does not do any defaulting.

This change allows the synthesizer to "fix itself" with a new release without being deadlocked, but does not address the defaulting, we might or might not add a feature for this.

AYM1607 (approved):

PR #184, opened 2024-08-01
Ignore nil input revisions

jveski:
It doesn't make sense to block synthesis when some inputs don't set a revision. If ordering isn't important, other controllers should be able to just omit the revisions.

### DIFF
--- a/internal/controllers/reconciliation/crud_test.go
+++ b/internal/controllers/reconciliation/crud_test.go
@@ -403,18 +403,7 @@ func TestReconcileInterval(t *testing.T) {
 	// Test subject
 	setupTestSubject(t, mgr)
 	mgr.Start(t)
-
-	// Any syn/comp will do since we faked out the synthesizer pod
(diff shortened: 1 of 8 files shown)


internal/controllers/reconciliation/crud_test.go
```diff
@@ -403,18 +403,7 @@ func TestReconcileInterval(t *testing.T) {
...
-	syn.Name = "test-syn"
-	syn.Spec.Image = "create"
-	require.NoError(t, upstream.Create(ctx, syn))
-
-	comp := &apiv1.Composition{}
-	comp.Name = "test-comp"
-	comp.Namespace = "default"
-	comp.Spec.Synthesizer.Name = syn.Name
-	require.NoError(t, upstream.Create(ctx, comp))
+	writeGenericComposition(t, upstream)
```
jveski, line 406: Bit of unrelated test refactoring

AYM1607 (approved):

PR #185, opened 2024-08-02
Helm hook manifest

chihshenghuang:
1. Append hook manifest data
2. Add withReleaseName option


jveski (approved):
LGTM! I don't know why it isn't showing the test status on this PR - I can see that the builds ran against your fork and passed. We should probably figure that out soon but I'm not worried about merging this first.

PR #186, opened 2024-08-05
Synth+input ordering

jveski:
Some inputs may be derived from annotations on synthesizer resources. This is actually quite a nice pattern for cases where external resources are fetched by another controller, and the set of resources fetched can be defined dynamically by the synthesizer. It becomes possible to request new resources by just updating the synthesizer object.

But currently it's possible that a newer version of the synthesizer would be used to synthesize an older (derived) input, which unnecessarily complicates the backcompat matrix. We can solve this by just blocking synthesis until the other controller has time to catch up.


AYM1607 (approved):

chihshenghuang:
LGTM!

PR #192, opened 2024-08-26
Better orphan'd resource cleanup

jveski:
Currently Eno will recreate missing namespaces in order to finish deleting any symphonies that still reference them. This behavior was symphony specific, which kind of makes sense given its role in the Eno ecosystem.

The problem is resources that _could in the future be_ related to a symphony, but are not yet. Specifically: resource slices. It's possible that synthesis was in-progress when the namespace was force-deleted, and that resource slices have been written but not yet associated with the composition.

I propose that we just enable the cleanup logic for all resources. There isn't any real reason for it to be exclusive to symphonies - just try to recreate the namespace any time it's still missing after 1 second.

### DIFF
--- a/docs/symphony.md
+++ b/docs/symphony.md
@@ -79,14 +79,3 @@ spec:
             name: a-different-input
             namespace: default
 ```
-
-## Deletion Behavior
-
-Symphonies are high-level resources designed to always converge, even in the face of rare split-brain states.
-
-Force deleting namespaces leaves resources in a strange state in which they exist but cannot be updated.
-Symphonies recover from this state by carefully recreating the namespace and forcibly removing internal finalizers.
-
-Because of this it's possible that managed resources will not be cleaned up if they exist outside of the orphaned namespace.
-Worst case, managed resources might be recreated by the Eno reconciler if it outpaces kube-controller-manager's namespace controller.
-Beware of this if you plan to use symphony resources to manage resources outside of the symphony's own namespace.
--- a/internal/controllers/liveness/namespace.go
+++ b/internal/controllers/liveness/namespace.go
@@ -23,10 +23,11 @@ var orphanableKinds = []string{"Symphony", "Composition", "ResourceSlice"}
 // This can happen if clients get tricky with the /finalize API.
 // Without this controller Eno resources will never be deleted since updates to remove the finalizers will fail.
 type namespaceController struct {
-	client client.Client
+	client              client.Client
+	creationGracePeriod time.Duration
 }
 
-func NewNamespaceController(mgr ctrl.Manager) error {
+func NewNamespaceController(mgr ctrl.Manager, creationGracePeriod time.Duration) error {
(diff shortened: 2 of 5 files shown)


internal/controllers/liveness/namespace.go
```diff
@@ -59,21 +70,27 @@ func (c *namespaceController) Reconcile(ctx context.Context, req ctrl.Request) (
 		logger.V(0).Info("deleting recreated namespace")
 		return ctrl.Result{}, nil
 	}
-	if err == nil { // important that this is the GET error
-		return ctrl.Result{}, nil // namespace exists, nothing to do
+	if err == nil {
```
AYM1607, line 73: What if this error is non-nil but not not-found?

internal/controllers/liveness/namespace.go
```diff
@@ -84,42 +101,47 @@ func (c *namespaceController) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	list.Kind = kind
+	list.APIVersion = "eno.azure.io/v1"
+	err := c.client.List(ctx, list, client.InNamespace(ns))
+	if err != nil {
+		return false, nil, fmt.Errorf("listing resources of kind %q: %w", kind, err)
+	}
+	if len(list.Items) == 0 {
+		return false, nil, nil // no orphaned resources, nothing to do
+	}
+	if delta := time.Since(mostRecentCreation(list)); delta < time.Second {
```
AYM1607, line 137: Should we make the delta threshold adjustable?

docs/symphony.md
```diff
@@ -79,14 +79,3 @@ spec:
             name: a-different-input
             namespace: default
 ```
-
```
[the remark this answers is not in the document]
jveski: That's interesting... I've been of the mind that the docs should be based on how setting various properties alters the behavior of Eno. So documenting behavior that isn't part of the API isn't really a thing yet. I'd say let's make a note to document this at some point, but I don't think it really fits anywhere currently - cool?

internal/controllers/liveness/namespace.go
jveski: lol That's completely reasonable - will do

AYM1607: Is 1 sec an informed guess? I'm unsure if etcd streaming for different Kinds could drift this much under different levels of load.

docs/symphony.md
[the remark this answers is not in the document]
AYM1607: Totally cool. I also don't like the idea of having a catch-all file that can potentially (more like surely?) grow unbounded. However, I believe it's worth at least mentioning it since it's a behavior affecting the whole cluster that wouldn't exist otherwise if Eno wasn't running

internal/controllers/liveness/namespace.go
jveski: That condition is handled right after the get call

AYM1607 (approved):

chihshenghuang (approved):

PR #193, opened 2024-08-27
Small improvements to the symphony implementation

jveski:
- Patch instead of update to avoid conflicts
- Deduplicate bindings by key
- Update composition when annotations change

### DIFF
--- a/internal/controllers/replication/symphony.go
+++ b/internal/controllers/replication/symphony.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"slices"
 	"sort"
+	"strings"
 
 	apiv1 "github.com/Azure/eno/api/v1"
 	"github.com/Azure/eno/internal/manager"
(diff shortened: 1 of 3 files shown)


chihshenghuang (approved):

internal/controllers/replication/symphony.go
```diff
@@ -173,7 +177,9 @@ func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.
 		// Diff and update if needed when the composition for this synthesizer already exists
 		if existings, ok := existingBySynthName[variation.Synthesizer.Name]; ok {
 			existing := existings[0]
-			if equality.Semantic.DeepEqual(comp.Spec, existing.Spec) && equality.Semantic.DeepEqual(comp.Labels, existing.Labels) {
+			if equality.Semantic.DeepEqual(comp.Spec, existing.Spec) &&
+				equality.Semantic.DeepEqual(comp.Labels, existing.Labels) &&
+				equality.Semantic.DeepEqual(comp.Annotations, existing.Annotations) {
```
AYM1607, line 182: Ahhh there was a reason why I didn't put this in the first time around, let me try to remember what it was

AYM1607: We need to update that code path to set the annotation in the variation instead...

PR #194, opened 2024-08-27
Merge symphony labels/annotations

jveski:
This ensures that the symphony controller doesn't stomp on any labels/annotations set outside of the symphony.


AYM1607 (approved):
LGTM

PR #195, opened 2024-09-05
Clean up Pods for compositions that no longer exist

AYM1607:
We should clean up Pods if compositions no longer exist. This scenario is possible if Comps are force deleted without honoring finalizers.

### DIFF
--- a/internal/controllers/synthesis/lifecycle.go
+++ b/internal/controllers/synthesis/lifecycle.go
@@ -58,7 +58,10 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 	logger := logr.FromContextOrDiscard(ctx)
 	comp := &apiv1.Composition{}
 	err := c.client.Get(ctx, req.NamespacedName, comp)
-	if err != nil {
+	if errors.IsNotFound(err) {
+		// Clean up Pods for composition that no longer exists.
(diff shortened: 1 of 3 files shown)


internal/controllers/synthesis/lifecycle.go
```diff
@@ -317,6 +320,30 @@ func shouldDeletePod(logger logr.Logger, comp *apiv1.Composition, syn *apiv1.Syn
...
+	for _, pod := range pods.Items {
+		if pod.DeletionTimestamp != nil {
+			continue
+		}
+		err := c.client.Delete(ctx, &pod)
+		if client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("deleting Pod %s: %w", pod.Name, err)
+		}
+		logger.V(0).Info("deleted synthesizer pod", "podName", pod.Name, "reason", "CompositionDoesNotExist")
+		return nil
```
chihshenghuang, line 342: why delete one pod only? should we delete all synthesizer pod if the comp doesn't exist?

AYM1607: This controller watches both Pods and Compositions, it should only do one write of the most granular resource per controller tick (the Pod in this case). The deletion of one Pod (either successful or unsuccessful) will trigger another controller tick and thus another deletion until we're done

jveski (approved):

PR #200, opened 2024-09-12
Implement SynthesisEnv

AYM1607:
`SynthesisEnv` is a list of env variables that are be made available to the synthesis Pod. They can be specified directly through a Composition spec or replicated from a Symphony to its managed Compositions.


api/v1/symphony.go
```diff
@@ -32,8 +32,13 @@ type SymphonySpec struct {
...
 	Variations []Variation `json:"variations,omitempty"`
 
-	// Bindings are inherited from all compositions managed by this symphony.
+	// Bindings are inherited by all compositions managed by this symphony.
 	Bindings []Binding `json:"bindings,omitempty"`
+
+	// SynthesisEnv
+	// Copied opaquely into the compositions managed by this symphony.
+	// +kubebuilder:validation:MaxItems:=500
+	SynthesisEnv []EnvVar `json:"synthesisEnv,omitempty"`
```
chihshenghuang, line 41: can we use EnvVar from "k8s.io/api/core/v1" directly and no need to defined new EnvVar in env.go?

AYM1607: I don't think it's appropriate to use it because that struct supports ValueFrom which allows references to ConfigMaps/Secrets. We're not (and probably won't for a long time) implementing ValueFrom so I don't think we should expose it in the API.

jveski (approved):
⚡

PR #201, opened 2024-09-12
Implement IgnoreSideEffects

AYM1607:
There are some scenarios in which it's useful to avoid synthesis on side effects. Changes to bound inputs and a composition's synthesizer fall under this category.

When a composition sets IgnoreSideEffects to true, only changes to the composition itself (or its parent Symphony) will trigger a resynthesis.

### DIFF
--- a/internal/controllers/rollout/integration_test.go
+++ b/internal/controllers/rollout/integration_test.go
@@ -100,7 +100,9 @@ func TestRolloutIgnoreSideEffects(t *testing.T) {
 	comp.Name = "test-comp"
 	comp.Namespace = "default"
 	comp.Spec.Synthesizer.Name = syn.Name
-	comp.Spec.IgnoreSideEffects = true
+	comp.Annotations = map[string]string{
+		"eno.azure.io/ignore-side-effects": "true",
+	}
 	require.NoError(t, cli.Create(ctx, comp))
 
 	// Initial creation.
@@ -126,7 +128,9 @@ func TestRolloutIgnoreSideEffects(t *testing.T) {
 
 	// Stop ignoring side effects.
 	compCopy := comp.DeepCopy()
-	comp.Spec.IgnoreSideEffects = false
+	comp.Annotations = map[string]string{
+		"eno.azure.io/ignore-side-effects": "false",
+	}
 	require.NoError(t, cli.Patch(ctx, comp, client.MergeFrom(compCopy)))
 
 	testutil.Eventually(t, func() bool {
--- a/internal/controllers/rollout/integration_test.go
+++ b/internal/controllers/rollout/integration_test.go
@@ -75,6 +75,81 @@ func TestSynthesizerRollout(t *testing.T) {
 	})
 }
 
+// TestRolloutIgnoreSideEffects proves that synthesizer changes are not rolled out to compositions which are ignoring side effects.
+func TestRolloutIgnoreSideEffects(t *testing.T) {
+	ctx := testutil.NewContext(t)
+	mgr := testutil.NewManager(t)
+	cli := mgr.GetClient()
+
+	require.NoError(t, flowcontrol.NewSynthesisConcurrencyLimiter(mgr.Manager, 10, 0))
+	require.NoError(t, NewSynthesizerController(mgr.Manager))
+	require.NoError(t, NewController(mgr.Manager, time.Millisecond*10))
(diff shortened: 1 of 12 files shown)


internal/controllers/rollout/integration_test.go
```diff
@@ -75,6 +75,81 @@ func TestSynthesizerRollout(t *testing.T) {
...
+	require.Less(t, comp.Status.CurrentSynthesis.ObservedSynthesizerGeneration, syn.Generation)
+
+	// Stop ignoring side effects.
+	compCopy := comp.DeepCopy()
+	comp.Spec.IgnoreSideEffects = false
+	require.NoError(t, cli.Patch(ctx, comp, client.MergeFrom(compCopy)))
+
+	testutil.Eventually(t, func() bool {
+		require.NoError(t, cli.Get(ctx, client.ObjectKeyFromObject(comp), comp))
+		return comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.ObservedSynthesizerGeneration >= syn.Generation
```
chihshenghuang, line 134: nit: use helper func isInSync https://github.com/Azure/eno/blob/9f0c1a8feaadcb0c59e6f543773518baa0e2308d/internal/controllers/rollout/synthesizer.go#L95

internal/controllers/rollout/integration_test.go
```diff
@@ -75,6 +75,81 @@ func TestSynthesizerRollout(t *testing.T) {
...
+			return err
+		}
+		syn.Spec.Image = "another-updated-image"
+		return cli.Update(ctx, syn)
+	})
+
+	// This time the rollout is observed.
+	testutil.Eventually(t, func() bool {
+		require.NoError(t, cli.Get(ctx, client.ObjectKeyFromObject(comp), comp))
+		return comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.ObservedSynthesizerGeneration >= syn.Generation
```
chihshenghuang, line 149: nit: use helper func isInSync https://github.com/Azure/eno/blob/9f0c1a8feaadcb0c59e6f543773518baa0e2308d/internal/controllers/rollout/synthesizer.go#L95

internal/controllers/watch/integration_test.go
```diff
@@ -79,6 +79,90 @@ func TestBasics(t *testing.T) {
...
+		initialResourceVersion = rv
+		return rv != ""
+	})
+
+	// Start to ignore side effects.
+	comp.Spec.IgnoreSideEffects = true
+	require.NoError(t, cli.Update(ctx, comp))
+
+	// Give some time to the controller to process this.
+	time.Sleep(time.Millisecond * 500)
```
chihshenghuang, line 137: wait short time might cause the unit test to be flaky, can we use testuitl.Eventually to check the comp.metadata.resourceVersion is changed after updating the spec to comp?

internal/controllers/rollout/integration_test.go
```diff
@@ -75,6 +75,81 @@ func TestSynthesizerRollout(t *testing.T) {
...
+	err := retry.RetryOnConflict(testutil.Backoff, func() error {
+		if err := cli.Get(ctx, client.ObjectKeyFromObject(syn), syn); err != nil {
+			return err
+		}
+		syn.Spec.Image = "updated-image"
+		return cli.Update(ctx, syn)
+	})
+	require.NoError(t, err)
+	// Give some time to the controller to process the change.
+	time.Sleep(time.Second)
```
chihshenghuang, line 122: wait short time might cause the unit test to be flaky, can we use testuitl.Eventually to check the syn.metadata.resourceVersion is changed after updating the spec.image to syn?

AYM1607: The issue here is that we're waiting for it to _not change_ so we can't use eventually. The only way that I can think of doing this is by somehow hooking into controller runtime and ensure that this write has been processed, but even then I don't know if it can be completely trusted. It's a good idea to explore other ways of testing negative cases though.

internal/controllers/watch/integration_test.go
AYM1607: Same as my other comment regarding delays.

internal/controllers/rollout/integration_test.go
chihshenghuang: i see, thanks for the explanation 👍

chihshenghuang (approved):
LGTM 🚀

jveski (approved):

PR #202, opened 2024-09-13
Add delete policy by using patch resource unit test

chihshenghuang:
Add unit test for the scenario that use the deletion patch to ensure the resource will be deleted before resource to be created or updated.

### DIFF
--- a/internal/controllers/reconciliation/patch_test.go
+++ b/internal/controllers/reconciliation/patch_test.go
@@ -3,7 +3,6 @@ package reconciliation
 import (
 	"context"
 	"testing"
-	"time"
 
 	apiv1 "github.com/Azure/eno/api/v1"
 	testv1 "github.com/Azure/eno/internal/controllers/reconciliation/fixtures/v1"
@@ -282,11 +281,7 @@ func TestPatchDeletionBeforeUpgrade(t *testing.T) {
 
 	// Create configmap first before the patch deletion is applied.
 	require.NoError(t, downstream.Create(ctx, cm))
-	creationTime := cm.GetCreationTimestamp()
-	creationResourceVersion := cm.GetResourceVersion()
-	// Wait for more than one second to ensure the createTimestamp is changed when recreating configmap,
-	// or it might have same createTimestamp and fail to pass test.
-	time.Sleep(1500 * time.Millisecond)
+	creationUID := cm.GetUID()
 
 	// Create deletion patch and configmap with new change.
 	_, comp := writeGenericComposition(t, upstream)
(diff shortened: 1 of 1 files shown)


internal/controllers/reconciliation/patch_test.go
```diff
@@ -127,3 +128,181 @@ func TestPatchDeletion(t *testing.T) {
...
+	setupTestSubject(t, mgr)
+	mgr.Start(t)
+
+	// Create configmap first before the patch deletion is applied.
+	require.NoError(t, downstream.Create(ctx, cm))
+	creationTime := cm.GetCreationTimestamp()
+	creationResourceVersion := cm.GetResourceVersion()
+	// Wait for more than one second to ensure the createTimestamp is changed when recreating configmap,
+	// or it might have same createTimestamp and fail to pass test.
+	time.Sleep(1500 * time.Millisecond)
```
jveski, line 289: Can we compare `ObjectMeta.UID` instead of timestamp to avoid slowing down the tests?

internal/controllers/reconciliation/patch_test.go
```diff
@@ -127,3 +128,181 @@ func TestPatchDeletion(t *testing.T) {
...
+	// Verify the configmap should be created after the deletion patch is applied.
+	err := downstream.Get(ctx, client.ObjectKeyFromObject(cm), cm)
+	require.NoError(t, err)
+	require.NotNil(t, cm)
+	require.Equal(t, val, cm.Data[key])
+}
+
+// TestPatchDeletionBeforeUpgrade proves that a patch resource can delete the resource it references before the resource is upgraded.
+// Basically, this is the same behavior as Helm hook event with delete policy "before-hook-creation".
+func TestPatchDeletionBeforeUpgrade(t *testing.T) {
```
jveski, line 213: Have you tested what happens if either the patch or the resource it deletes are reconciled again after initial reconciliation? i.e. set the reconciliation interval annotation to ~1ms and somehow prove that it doesn't churn?

I suspect that every time the configmap is reconciled it will be recreated, which will trigger reconciliation of the patch since it's in a future readiness group, which will delete it. Not 100% sure though.

internal/controllers/reconciliation/patch_test.go
chihshenghuang: good idea! update test to use uid for validation.

internal/controllers/reconciliation/patch_test.go
chihshenghuang: discuss offline, will test this scenario for the resource reconciliation to see the behavior of deletion patch in next PR

jveski (approved):

PR #203, opened 2024-09-13
Log synthesis ID

keith-ms:
There exists a unique synthesizer identifier for each synthesis that occurs, but it isn't currently logged. This change adds the synthesis ID to log output where it exists.

Ran unit tests here:
https://github.com/keith-ms/eno/actions/runs/10853836588


keith-ms:
@microsoft-github-policy-service agree company="Microsoft"

internal/controllers/aggregation/slice.go
```diff
@@ -36,7 +36,15 @@ func (s *sliceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	if err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("getting composition: %w", err))
 	}
-	logger = logger.WithValues("compositionGeneration", comp.Generation, "compositionName", comp.Name, "compositionNamespace", comp.Namespace)
+
+	synthesisID := ""
+	if comp.Status.CurrentSynthesis != nil {
```
jveski, line 41: Would you mind moving this logic to a getter-style func on the composition status struct? To avoid duplicating the logic across controllers.

Like: `func (s *CompositionStatus) GetCurrentSynthesisUUID() string {`

jveski (approved):

internal/controllers/aggregation/slice.go
jveski: (just a nit, feel free to merge without addressing)

keith-ms: Updated.

jveski (approved):
:ship:

PR #204, opened 2024-09-19
Add reconcile interval test

chihshenghuang:
1. Resource with interval won't be re-created after deletion by patch
2. Resource with interval won't trigger patch to delete resource


internal/controllers/reconciliation/patch_test.go
```diff
@@ -299,3 +300,164 @@ func TestPatchDeletionBeforeUpgrade(t *testing.T) {
...
+	testutil.WithFakeExecutor(t, mgr, func(ctx context.Context, s *apiv1.Synthesizer, input *krmv1.ResourceList) (*krmv1.ResourceList, error) {
+		obj := &unstructured.Unstructured{
+			Object: map[string]any{
+				"apiVersion": "eno.azure.io/v1",
+				"kind":       "Patch",
+				"metadata": map[string]any{
+					"name":      cmName,
+					"namespace": cmNamespace,
+					"annotations": map[string]string{
+						// This patch should be applied before the configmap is re-created.
```
chihshenghuang, line 337: reconcile-interval annotation won't work for patch if i understand it correctly, it won't re-queue due to the resource.Deleted() always be true for deletion patch https://github.com/Azure/eno/blob/c2af2ca0b63171213a0f98aa2a7c3b91f70000f0/internal/resource/resource.go#L63

internal/controllers/reconciliation/patch_test.go
```diff
@@ -299,3 +300,164 @@ func TestPatchDeletionBeforeUpgrade(t *testing.T) {
...
+	require.NoError(t, err)
+
+	uid := cm.GetUID()
+	// Ensure the configmap is reconciled at least once.
+	time.Sleep(100 * time.Millisecond)
+	err = downstream.Get(ctx, client.ObjectKeyFromObject(cm), cm)
+	require.NoError(t, err)
+
+	newUID := cm.GetUID()
+	// Verify the configmap is not re-created.
```
chihshenghuang, line 461: the configmap with higher readiness group will be re-queue and deletion patch is delivered before confimap create or and reconciled, so the configmap won't be deleted. Unless the synthesizer pod create and apply the deletion patch to delete existing configmap.

internal/controllers/reconciliation/patch_test.go
jveski: Wow this is not what I expected at all 😆 . I actually think this behavior makes sense, but we should probably document it somewhere, at some point. Since it does kind of violate the semantics of the reconcile-interval annotation.

internal/controllers/reconciliation/patch_test.go
```diff
@@ -299,3 +300,164 @@ func TestPatchDeletionBeforeUpgrade(t *testing.T) {
...
+
+		return &krmv1.ResourceList{Items: []*unstructured.Unstructured{obj}}, nil
+	})
+
+	// Test subject
+	setupTestSubject(t, mgr)
+	mgr.Start(t)
+
+	// Create configmap first before the patch deletion is applied.
+	require.NoError(t, downstream.Create(ctx, cm))
```
jveski, line 359: We should create `cm` with Eno instead of directly with client-go - Eno only "knows about" the reconcile-interval of resources it manages.

jveski: Actually- maybe we can just remove this test? `TestPatchDeletionForResourceWithReconciliationFromInput` covers the scenario I had in mind

chihshenghuang: i was trying to check the behavior of applying the patch deletion to the existing resource, but you are right that we should test for the resource created by eno only. Remove this test.

jveski (approved):

PR #206, opened 2024-09-30
fixing time to always use abs and postive time stamps

grantmanleymicrosoft:
Fixing the time que for latency to always report positive latency items

### DIFF
--- a/internal/controllers/aggregation/slice.go
+++ b/internal/controllers/aggregation/slice.go
@@ -98,8 +98,8 @@ func (s *sliceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 		if synthed := comp.Status.CurrentSynthesis.Synthesized; synthed != nil {
 			latency := maxReadyTime.Sub(synthed.Time)
-			if latency.Microseconds() > 0 {
-				logger.V(0).Info("composition became ready", "latency", latency.Abs().Microseconds())
+			if latency.Milliseconds() > 0 {
+				logger.V(0).Info("composition became ready", "latency", latency.Abs().Milliseconds())
 			}
 		}
 	} else {
@@ -111,7 +111,7 @@ func (s *sliceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 		if synthed := comp.Status.CurrentSynthesis.Synthesized; synthed != nil {
 			latency := now.Sub(synthed.Time)
-			logger.V(0).Info("composition was reconciled", "latency", latency.Abs().Microseconds())
+			logger.V(0).Info("composition was reconciled", "latency", latency.Abs().Milliseconds())
 		}
 	} else {
 		comp.Status.CurrentSynthesis.Reconciled = nil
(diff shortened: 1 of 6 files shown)


grantmanleymicrosoft:
@microsoft-github-policy-service agree company="Microsoft"

internal/controllers/aggregation/slice.go
```diff
@@ -98,8 +98,8 @@ func (s *sliceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 		if synthed := comp.Status.CurrentSynthesis.Synthesized; synthed != nil {
 			latency := maxReadyTime.Sub(synthed.Time)
-			if latency.Milliseconds() > 0 {
-				logger.V(0).Info("composition became ready", "latency", latency.Milliseconds())
+			if latency.Microseconds() > 0 {
+				logger.V(0).Info("composition became ready", "latency", latency.Abs().Microseconds())
```
jveski, line 102: Any particular reason for switching to microseconds? Milliseconds is more... intuitive or at least common IMO.

Also- I'm curious what `Abs` does in this case, since we're already checking for positive.

internal/controllers/aggregation/slice.go
```diff
@@ -111,15 +111,17 @@ func (s *sliceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
-			logger.V(0).Info("composition was reconciled", "latency", latency.Milliseconds())
+			logger.V(0).Info("composition was reconciled", "latency", latency.Abs().Microseconds())
 		}
 	} else {
 		comp.Status.CurrentSynthesis.Reconciled = nil
 	}
 
 	err = s.client.Status().Update(ctx, comp)
 	if err != nil {
+		logger.V(0).Info(" Error in aggregating resource status into composition", "msg", err)
```
jveski, line 122: I have a resistance to logging like this since it ends up with duplicate entries (the manager will log any returned errors as well).

Kind of related: I'd love to find some way to inject extra properties (compositionName, etc.) into the error logs emitted by the manager since right now they are sometimes lacking context.

internal/controllers/aggregation/slice.go
grantmanleymicrosoft: Fixed it so that we only use milliseconds

internal/controllers/aggregation/slice.go
grantmanleymicrosoft: Let me know if this is what you were looking for , I added extra logging with , return ctrl.Result{}, fmt.Errorf("updating composition '%s' status: %w", comp.Name, err)

jveski: That works for now. Ultimately I want to find a way to inject the fields from the controller itself, so they don't have to be added by every controller. But there are some perf concerns since it requires hitting the informer cache (IIRC), so I've been putting it off.

In any case, thanks for addressing this error - I'll make an issue for the fancy fields injection work.

jveski (approved):

PR #207, opened 2024-09-30
Fix ignore side effects

AYM1607:
Fixes a bug where a composition is never synthesized if it was created ignoring side effects.

### DIFF
--- a/examples/ignore-side-effects/main.go
+++ b/examples/ignore-side-effects/main.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"strconv"
+
+	"github.com/Azure/eno/pkg/function"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+func main() {
+	w := function.NewDefaultOutputWriter()
+	r, err := function.NewDefaultInputReader()
+	if err != nil {
+		panic(err) // non-zero exits will be retried
+	}
+
+	input := &corev1.ConfigMap{}
+	function.ReadInput(r, "example-input", input)
+
+	replicas, _ := strconv.Atoi(input.Data["replicas"])
+
+	deploy := &appsv1.Deployment{}
+	deploy.APIVersion = "apps/v1"
+	deploy.Kind = "Deployment"
+	deploy.Name = "example-nginx-deployment"
+	deploy.Namespace = "default"
+	deploy.Spec.Replicas = ptr.To(int32(replicas))
+	deploy.Spec.Selector = &metav1.LabelSelector{MatchLabels: map[string]string{"app": "nginx-example"}}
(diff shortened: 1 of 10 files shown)


internal/controllers/synthesis/lifecycle.go
```diff
@@ -363,12 +363,17 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
...
-	// - the bound input resources have changed
+	// - a side effect that has not observed a synthesis has occurred
+	//		The side effects observed by this controller are:
+	//			- changes to non-defferred inputs.
 	// AND
 	// - synthesis is not already pending
 	// - all bound input resources exist and are in lockstep (or composition is being deleted)
 	syn := comp.Status.CurrentSynthesis
-	return (comp.DeletionTimestamp != nil || (comp.InputsExist(synth) && !comp.InputsMismatched(synth))) && (syn == nil || syn.ObservedCompositionGeneration != comp.Generation || (!inputRevisionsEqual(synth, comp.Status.InputRevisions, syn.InputRevisions) && syn.Synthesized != nil))
+	return (syn == nil ||
```
jveski, line 373: So many parens it's starting to look like lisp 🥇 

Nice refactor, this was... not beautiful before.

internal/controllers/rollout/controller.go
```diff
@@ -65,6 +65,20 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// Do this before calculating selecting the next composition to avoid
+	// skewing the results.
+	for _, comp := range comps.Items {
+		if comp.Status.PendingResynthesis != nil && comp.ShouldIgnoreSideEffects() {
+			comp.Status.PendingResynthesis = nil
+			err := c.client.Status().Update(ctx, &comp)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("clearing PendingResythesis field for %s in namespace %s: %w", comp.Name, comp.Namespace, err)
+			}
+			return ctrl.Result{Requeue: true}, nil
```
jveski, line 78: Curious why we need to requeue here?

internal/controllers/synthesis/lifecycle.go
AYM1607: It'll need more work, maybe in another PR...

internal/controllers/rollout/controller.go
AYM1607: Good catch, we don't, we can rely on the write to kick off the next tick

examples/ignore-side-effects/main.go
```diff
@@ -0,0 +1,46 @@
+package main
```
jveski, line 1: nit: would prefer a toy shell script synth to avoid an extra image build during the smoke tests

jveski (approved):

examples/ignore-side-effects/main.go
AYM1607: I'll track this in #208

PR #210, opened 2024-10-02
Avoid assignments to nil maps when coalescing metadata

AYM1607:
Initialize the maps when they're nil.

### DIFF
--- a/internal/controllers/replication/symphony.go
+++ b/internal/controllers/replication/symphony.go
@@ -265,13 +265,21 @@ func sortSynthesizerRefs(refs []apiv1.SynthesizerRef) {
 func coalesceMetadata(variation *apiv1.Variation, existing *apiv1.Composition) bool {
 	var metaChanged bool
 	for key, val := range variation.Labels {
-		if existing.Labels == nil || existing.Labels[key] != val {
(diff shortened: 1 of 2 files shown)


chihshenghuang (approved):

internal/controllers/replication/symphony.go
```diff
@@ -265,13 +265,21 @@ func sortSynthesizerRefs(refs []apiv1.SynthesizerRef) {
 func coalesceMetadata(variation *apiv1.Variation, existing *apiv1.Composition) bool {
 	var metaChanged bool
 	for key, val := range variation.Labels {
-		if existing.Labels == nil || existing.Labels[key] != val {
+		if existing.Labels == nil {
+			existing.Labels = map[string]string{}
+			metaChanged = true
```
jveski, line 270: This is kind of a nit, but I think we should avoid considering the initial allocation a change. Since apiserver uses omitEmpty it won't return an empty map, so it's possible to get stuck in a loop here if we for some reason didn't set a key.

Of course given the code directly below this isn't possible, but I worry about future changes, etc.

jveski (approved):

PR #212, opened 2024-10-07
Fix resource slice race condition

jveski:
It's possible that the reconciler process "sees" a resource slice before the corresponding composition. Currently we handle this correctly after the initial synthesis by comparing generations. But the logic is inverted in the case that the synthesis is nil - __resource slices are deleted if there is a newer composition generation__.

For example:

- Composition is created
- Composition hits the reconciler informers
- Synthesis starts and completes
- Resource slice hits the reconciler informers
  - The slice is deleted by `shouldDeleteSlice` - either because the composition has been updated (newer generation than resource slice), or because composition is deleted)
- Synthesis status hits the reconciler informers

This doesn't break tests because the integration tests use the same informers for some controllers that normally run in separate processes, so events are received in the same order by both sets of controllers.

### DIFF
--- a/internal/controllers/synthesis/slicecleanup.go
+++ b/internal/controllers/synthesis/slicecleanup.go
@@ -98,9 +98,9 @@ func shouldDeleteSlice(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool
 	}
 	isOutdated := slice.Spec.Attempt != 0 && comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Attempts > slice.Spec.Attempt
 	isReferencedByComp := synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) || synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice)
-	isPendingSynthesis := comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Synthesized != nil
+	isSynthesized := comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Synthesized != nil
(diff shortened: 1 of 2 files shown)


internal/controllers/synthesis/slicecleanup.go
```diff
@@ -93,23 +93,23 @@ func (c *sliceCleanupController) Reconcile(ctx context.Context, req ctrl.Request
...
 
 func shouldDeleteSlice(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool {
-	if comp.Status.CurrentSynthesis != nil && slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration {
+	if comp.Status.CurrentSynthesis == nil || slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration {
 		return false // stale informer
 	}
 	isOutdated := slice.Spec.Attempt != 0 && comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Attempts > slice.Spec.Attempt
 	isReferencedByComp := synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) || synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice)
-	isPendingSynthesis := comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Synthesized != nil
+	isSynthesized := comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Synthesized != nil
```
jveski, line 101: Renamed for clarity

chihshenghuang (approved):

PR #214, opened 2024-10-08
Fix input revision comparison when some inputs are deferreed

jveski:
Currently deferred inputs will cause `equal` to be > len(a)

### DIFF
--- a/internal/controllers/synthesis/lifecycle.go
+++ b/internal/controllers/synthesis/lifecycle.go
@@ -413,16 +413,15 @@ func inputRevisionsEqual(synth *apiv1.Synthesizer, a, b []apiv1.InputRevisions) 
 	sort.Slice(b, func(i, j int) bool { return b[i].Key < b[j].Key })
 
 	var equal int
-	for _, ar := range a {
-		for _, br := range b {
-			if ref, exists := refsByKey[ar.Key]; exists && ref.Defer {
-				equal++
-				continue // ignore deferred inputs
-			}
+	for i, ar := range a {
+		br := b[i]
+		if ref, exists := refsByKey[ar.Key]; exists && ref.Defer {
+			equal++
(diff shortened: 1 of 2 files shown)


internal/controllers/synthesis/lifecycle.go
```diff
@@ -413,16 +413,15 @@ func inputRevisionsEqual(synth *apiv1.Synthesizer, a, b []apiv1.InputRevisions)
 	sort.Slice(b, func(i, j int) bool { return b[i].Key < b[j].Key })
 
 	var equal int
-	for _, ar := range a {
-		for _, br := range b {
-			if ref, exists := refsByKey[ar.Key]; exists && ref.Defer {
```
AYM1607, line 418: Ahhhh, each defer input would add to `equal` once per key in `a`

internal/controllers/synthesis/lifecycle_test.go
```diff
@@ -674,14 +674,75 @@ func TestShouldSwapStates(t *testing.T) {
...
+			equal: true,
+		},
+		{
+			name:  "mixed resource version and revision",
+			a:     []apiv1.InputRevisions{{Key: "foo", ResourceVersion: "not-zero"}, {Key: "baz"}},
+			b:     []apiv1.InputRevisions{{Key: "foo", Revision: ptr.To(123)}, {Key: "baz"}},
+			equal: false,
+		},
+		{
+			name:  "ignore deferred",
```
AYM1607, line 727: This one fails with the old logic, we missed it because the tests were only comparing 1 revisions struct for deferred scenarios which masked the bug

jveski: 🤦 🤦 🤦

AYM1607 (approved):

PR #217, opened 2024-10-11
Handle kubelet partitions

jveski:
Sadly, k8s doesn't have a good mechanism for handling new pods on nodes that recently updated their lease (are still Ready) before being partitioned from apiserver. Pods will sit around until the node becomes NotReady and is eventually reflected in the pod status.

Synthesis tail latency suffers in this case so it's worth implementing our own timeouts on the relevant phase of the pod lifecycle: the window between being scheduled and ack'd by kubelet.

### DIFF
--- a/cmd/eno-controller/main.go
+++ b/cmd/eno-controller/main.go
@@ -65,6 +65,7 @@ func runController() error {
 	flag.StringVar(&synconf.PodNamespace, "synthesizer-pod-namespace", os.Getenv("POD_NAMESPACE"), "Namespace to create synthesizer pods in. Defaults to POD_NAMESPACE.")
 	flag.StringVar(&synconf.ExecutorImage, "executor-image", os.Getenv("EXECUTOR_IMAGE"), "Reference to the image that will be used to execute synthesizers. Defaults to EXECUTOR_IMAGE.")
 	flag.StringVar(&synconf.PodServiceAccount, "synthesizer-pod-service-account", "", "Service account name to be assigned to synthesizer Pods.")
+	flag.DurationVar(&synconf.ContainerCreationTimeout, "container-creation-ttl", time.Second*3, "Timeout when waiting for kubelet to ack scheduled pods. Protects tail latency from kubelet network partitions")
 	flag.BoolVar(&debugLogging, "debug", true, "Enable debug logging")
 	flag.DurationVar(&watchdogThres, "watchdog-threshold", time.Minute*5, "How long before the watchdog considers a mid-transition resource to be stuck")
 	flag.DurationVar(&rolloutCooldown, "rollout-cooldown", time.Minute, "How long before an update to a related resource (synthesizer, bindings, etc.) will trigger a second composition's re-synthesis")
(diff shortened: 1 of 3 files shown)


cmd/eno-controller/main.go
```diff
@@ -65,6 +65,7 @@ func runController() error {
 	flag.StringVar(&synconf.PodNamespace, "synthesizer-pod-namespace", os.Getenv("POD_NAMESPACE"), "Namespace to create synthesizer pods in. Defaults to POD_NAMESPACE.")
 	flag.StringVar(&synconf.ExecutorImage, "executor-image", os.Getenv("EXECUTOR_IMAGE"), "Reference to the image that will be used to execute synthesizers. Defaults to EXECUTOR_IMAGE.")
 	flag.StringVar(&synconf.PodServiceAccount, "synthesizer-pod-service-account", "", "Service account name to be assigned to synthesizer Pods.")
+	flag.DurationVar(&synconf.ContainerCreationTimeout, "container-creation-ttl", time.Second*3, "Timeout when waiting for kubelet to ack scheduled pods. Protects tail latency from kubelet network partitions")
```
chihshenghuang, line 68: will 3 seconds too short for kubelet to schedule pod? curious how to define this value used here

jveski: The timeout only covers the period between when the pod is scheduled and the container statuses are written by kubelet, which normally takes <10ms.

But setting it to something reasonable like 100ms isn't safe since eno-controller and kube-scheduler probably aren't running on the same node so clock skew might be enough to make every pod timeout. So 3s seems like a reasonable tradeoff.

If nodes have clock skew of >3s, Eno will delete the first pod but won't delete the second until the first pod has had time to terminate, which takes a while. So even extreme clock skew is relatively safe.

One other safeguard we could add: disable this timeout after a few retries. That would handle very high clock skew _and_ fast pod termination (unlikely, but theoretically possible). Thoughts?

jveski: Now that I think about it, that other safeguard sounds like a good idea - just pushed a commit :)

internal/controllers/synthesis/lifecycle.go
```diff
@@ -303,14 +308,28 @@ func shouldDeletePod(logger logr.Logger, comp *apiv1.Composition, syn *apiv1.Syn
...
+		// Delete pods if they have been scheduled but not picked up by that node's kubelet
+		// This can happen when the node is Ready but recently partitioned from apiserver
+		//
+		// Clock jitter is a risk since the scheduled timestamp is written by the scheduler
+		// So we only enforce this timeout when a new pod can be created immediately
+		// i.e. when another pod for this synthesis isn't already terminating
+		// AND we bail out when the synthesis has already been tried a few times (what's a few more seconds latency at that point)
+		seenByKubelet := len(pod.Status.ContainerStatuses) != 0
+		retryPressure := comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Attempts > 3
+		if scheduledTime := getPodScheduledTime(&pod); !onePodDeleting && !seenByKubelet && !retryPressure && scheduledTime != nil && time.Since(*scheduledTime) > creationTTL {
```
AYM1607, line 328: Nice little conditional chain here 😄

AYM1607 (approved):

PR #235, opened 2024-10-24
Fix reconciler composition selector flags

jveski:
eno-reconciler processes that use selectors to restrict compositions they are responsible for reconciling will happily delete all resource slices __not__ matched by the selector - not a great failure mode lol.

The fix for `--composition-namespace` is easy - only watch resource slices in the composition's namespace. But `--composition-label-selector` is tricky since it matches on _composition_ labels, which don't necessarily match resource slices.

The simplest fix is to move the slice cleanup controller from the reconciler to the eno-controller process so it will always have a coherent view of the entire cluster. Historically, the controller/reconciler divide was designed such that the controller doesn't need to watch/cache resource slices, so the downside of this change is added overhead from the new informer. But practically speaking it just isn't an issue (we're talking about tens of MBs of memory worst case).

This change also adds an extra check to the resource slice cleanup controller to avoid any potential issues caused by stale informers: re-run the cleanup logic against a non-cached version of the composition. Since resource slices are so important to Eno, added safety justifies the overhead of a few get requests to apiserver.

### DIFF
--- a/cmd/eno-reconciler/main.go
+++ b/cmd/eno-reconciler/main.go
@@ -13,10 +13,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	ctrl "sigs.k8s.io/controller-runtime"
 
-	"github.com/Azure/eno/internal/controllers/aggregation"
 	"github.com/Azure/eno/internal/controllers/liveness"
 	"github.com/Azure/eno/internal/controllers/reconciliation"
-	"github.com/Azure/eno/internal/controllers/synthesis"
 	"github.com/Azure/eno/internal/flowcontrol"
 	"github.com/Azure/eno/internal/k8s"
 	"github.com/Azure/eno/internal/manager"
@@ -88,16 +86,6 @@ func run() error {
 	mgr, err := manager.NewReconciler(logger, mgrOpts)
 	if err != nil {
 		return fmt.Errorf("constructing manager: %w", err)
-	}
-
-	err = aggregation.NewSliceController(mgr)
-	if err != nil {
-		return fmt.Errorf("constructing status aggregation controller: %w", err)
-	}
-
-	err = synthesis.NewSliceCleanupController(mgr)
-	if err != nil {
-		return fmt.Errorf("constructing resource slice cleanup controller: %w", err)
 	}
 
 	if namespaceCleanup {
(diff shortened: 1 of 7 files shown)


internal/controllers/synthesis/slicecleanup_test.go
```diff
@@ -365,3 +365,36 @@ func TestShouldDeleteSlice(t *testing.T) {
 		})
 	}
 }
+
+func TestBuildCleanupDecision_StaleCache(t *testing.T) {
```
jveski, line 369: The other codepaths are covered by existing integration tests

cmd/eno-reconciler/main.go
```diff
@@ -90,16 +88,6 @@ func run() error {
 		return fmt.Errorf("constructing manager: %w", err)
 	}
 
-	err = aggregation.NewSliceController(mgr)
```
AYM1607, line 93: So the reconciler is now only in charge syncing slices into real resources?

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -39,36 +43,20 @@ func (c *sliceCleanupController) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	owner := metav1.GetControllerOf(slice)
-
-	// We only get the composition if it exists
-	// It shouldn't be possible that it doesn't exist, but still worth handling in case anyone creates an ad-hoc resource slice for some reason (it won't do anything tho)
```
AYM1607, line 44: Apparently it was very possible 😭

cmd/eno-reconciler/main.go
jveski: Exactly, probably more intuitive anyway

internal/controllers/synthesis/slicecleanup.go
jveski: I should tally up all of my comments like this that ended up being removed in bugfix PRs 😆

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -81,17 +69,83 @@ func (c *sliceCleanupController) Reconcile(ctx context.Context, req ctrl.Request
...
+}
+
+type cleanupDecision struct {
+	DoNotDelete   bool
+	HoldFinalizer bool
+	DeferBy       *time.Duration
+}
+
+func (c *cleanupDecision) String() string {
+	return fmt.Sprintf("DoNotDelete=%t,HoldFinalizer=%t", c.DoNotDelete, c.HoldFinalizer)
```
AYM1607, line 146: No defer by?

AYM1607 (approved):

internal/controllers/synthesis/slicecleanup.go
AYM1607: Yeah, pragmatically is not relevant, just odd to see a field missing without knowing the whole context, will leave it up to you

PR #240, opened 2024-11-01
Retry batched resource slice writes more aggressively

### LINKED ISSUE #226: More aggressive writebuffer behavior
The write buffer is tuned to be very cautious about overloading apiserver at the cost of (eno) latency.

Instead of waiting for the entire flush interval to retry errors we should use the standard workqueue exponential backoff rate limiter.

jveski:
Resolves #226

It doesn't make sense to wait the full batch interval before retrying errors since it's unlikely that the full cost of the request was paid by apiserver if it returned an error. Essentially, the value of the batch interval is reducing load at the cost of latency. But in most cases where apiserver returns an error we don't get the benefit of reduced load, only the cost of high latency.

We can improve tail latency by just using reasonable exponential backoff for the first few retries, then falling back to the batch interval after that for safety. Since the write buffer forget items on success the backoff logic only applies to errors, not requeues caused by changes enqueued concurrent with the last batch.

Also adds some tooling to help uncover issues like this in the future - optional error injection at the client level.

### DIFF
--- a/internal/flowcontrol/writebuffer.go
+++ b/internal/flowcontrol/writebuffer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/flowcontrol/writebuffer.go
```diff
@@ -207,3 +208,44 @@ type jsonPatch struct {
...
+	if failures > 0 && failures < 5 {
+		return time.Duration(10*math.Pow(2, float64(failures-1))) * time.Millisecond
+	}
+
+	// Non-error batching interval
+	return r.limiter.Reserve().Delay()
+}
+
+func (r *rateLimiter) NumRequeues(item interface{}) int {
+	r.failuresLock.Lock()
```
AYM1607, line 242: RWMutex?

jveski: NumRequeues shouldn't actually be called anywhere since nothing wraps this rate limiter impl. And forget is called from the same goroutine as When. So I don't think there's really a perf upside

AYM1607 (approved):

PR #241, opened 2024-11-04
Optionally block patches until composition deletion

### LINKED ISSUE #232: Patch hook for composition deletion
It should be possible to delete resources using the patch pseudo-resource only once the composition has been deleted. This is useful for cleaning up after controllers.

jveski:
Resolves #232

There's a common use case in which Eno deploys a controller which then creates some resources that aren't managed by Eno. So if Eno removes the controller, those resources will never be cleaned up.

An easy way to handle this: allow patches to be deferred until the composition is deleted. Deletion patches can then be defined to remove any known state that might be left behind by the controller.

### DIFF
--- a/docs/advanced-synthesis.md
+++ b/docs/advanced-synthesis.md
@@ -55,3 +55,20 @@ patch:
   ops:
     - { "op": "add", "path": "/metadata/deletionTimestamp", "value": "anything" }
 ```
+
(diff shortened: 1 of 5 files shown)


chihshenghuang:
🚢

chihshenghuang (approved):

PR #242, opened 2024-11-05
Fix patch edgecases

jveski:
Values that change on the server side should not cause the patch to be continuously applied. There are two known cases where this can occur: bools explicitly set to false, and resource quantities.

### DIFF
--- a/internal/resource/resource.go
+++ b/internal/resource/resource.go
@@ -1,6 +1,7 @@
 package resource
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


internal/resource/resource.go
```diff
@@ -19,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/kubectl/pkg/scheme"
```
AYM1607, line 23: Sadness 😞

AYM1607 (approved):
I think this is perfectly reasonable to unblock our current efforts, but we should create an issue to track an implementation of a dynamic runtime approach...

PR #244, opened 2024-11-13
Log the reason why swap comp synthesis

### LINKED ISSUE #225: Log reason why state was swapped
A composition's current synthesis is "swapped" to a new one when it's time to resynthesize due to an input/composition/synthesizer change. Currently it's hard to tell why this occurred since it requires correlating between log messages for a given composition. We should log a `reason` field, similar to pod deletion.

chihshenghuang:
Resolve #225 

Log the reason why swap comp synthesis during reconciliation process

### DIFF
--- a/internal/controllers/synthesis/lifecycle.go
+++ b/internal/controllers/synthesis/lifecycle.go
@@ -421,9 +421,9 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) (string
 
 	// Second condition
 	if isCompDeleted {
-		reason += "&& CompositionDeleted"
+		reason += " && CompositionDeleted"
 	} else if isCompInputsExist && !isCompInputsOutOfLockstep {
-		reason += "&& InputsInLockstep"
+		reason += " && InputsInLockstep"
 	}
 
 	if !(isCompDeleted || (isCompInputsExist && !isCompInputsOutOfLockstep)) {
--- a/internal/controllers/synthesis/lifecycle.go
+++ b/internal/controllers/synthesis/lifecycle.go
@@ -139,8 +139,9 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	// Swap the state to prepare for resynthesis if needed
-	logger, shouldSwap := shouldSwapStates(logger, syn, comp)
+	reason, shouldSwap := shouldSwapStates(logger, syn, comp)
 	if shouldSwap {
+		logger = logger.WithValues("reason", reason)
 		SwapStates(comp)
 		if err := c.client.Status().Update(ctx, comp); err != nil {
 			return ctrl.Result{}, fmt.Errorf("swapping compisition state: %w", err)
@@ -382,7 +383,7 @@ func SwapStates(comp *apiv1.Composition) {
 	}
 }
 
-func shouldSwapStates(logger logr.Logger, synth *apiv1.Synthesizer, comp *apiv1.Composition) (logr.Logger, bool) {
+func shouldSwapStates(logger logr.Logger, synth *apiv1.Synthesizer, comp *apiv1.Composition) (string, bool) {
 	// synthesize when (either):
 	// - synthesis has never occurred
 	// - the spec has changed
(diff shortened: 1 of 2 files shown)


internal/controllers/synthesis/lifecycle.go
```diff
@@ -392,10 +393,46 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
...
+	isCompGenDiff := syn != nil && syn.ObservedCompositionGeneration != comp.Generation
+	isInputRevisionsEqual := syn != nil && inputRevisionsEqual(synth, comp.Status.InputRevisions, syn.InputRevisions)
+	isSynthesizedNotNil := syn != nil && syn.Synthesized != nil
+	shouldIgnoreSideEffects := comp.ShouldIgnoreSideEffects()
+	isCompDeleted := comp.DeletionTimestamp != nil
+	isCompInputsExist := comp.InputsExist(synth)
+	isCompInputsOutOfLockstep := comp.InputsOutOfLockstep(synth)
+
+	reason := ""
+	if isSynNil {
```
jveski, line 407: We don't need to accumulate all of the reasons, it's fine to just return the first one since they're already evaluated in a sensible order.

That can simplify the implementation a bit- you can perform each check and return on failure like:

```go
if !comp.InputsExist(synth) {
	return logger.WithValues(...), false
}
```

internal/controllers/synthesis/lifecycle.go
```diff
@@ -392,10 +393,46 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
...
+		reason += ", composition inputs exist and inputs are not out of lockstep"
+	}
+
+	res := (isSynNil ||
+		isCompGenDiff ||
+		(!isInputRevisionsEqual && isSynthesizedNotNil && !shouldIgnoreSideEffects)) &&
+		(isCompDeleted || (isCompInputsExist && !isCompInputsOutOfLockstep))
+
+	if res {
+		logger = logger.WithValues("swap composition synthesis reason", reason)
```
jveski, line 432: Let's follow the pattern used for synthesis pod deletion where the column is "reason" and the value is a CamelCase description like `InputsMissing`

internal/controllers/synthesis/lifecycle.go
```diff
@@ -392,10 +393,46 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
...
+	}
+
+	return logger, res
```
jveski, line 435: nit: Consider returning a string instead of wrapping the logger, easier to unit test

internal/controllers/synthesis/lifecycle.go
chihshenghuang: The final result is calculated by AND first and second condition, and there are multiple OR condition inside first/second condition, so it can't tell the final result by checking only one condition.

From the example you mentioned, need to check "comp.DeletionTimestamp == nil" also to know the final result is false
```go
# second condition
(comp.DeletionTimestamp != nil || (comp.InputsExist(synth) && !comp.InputsOutOfLockstep(synth)))
```

I made the change to log the each result meet the condition by order instead of accumulate all the result meet the condition, but it's difficult to separate it by log only one condition, see the example log in next comment. Any better idea?

internal/controllers/synthesis/lifecycle.go
chihshenghuang: The reason to swap the comp is decided from multiple condition and I try to minimize the message, example reason: "CompositionGenerationDiff, CompositionInputsExist && InputsNotOutOfLockstep", is the format ok to follow the same pattern?

internal/controllers/synthesis/lifecycle.go
chihshenghuang: done

internal/controllers/synthesis/lifecycle.go
jveski: Ahh I see the problem - we need to invert the logic

For example:

```go
if syn.ObservedCompositionGeneration != comp.Generation {
  return "CompositionModified", true
}

if !inputRevisionsEqual(synth, comp.Status.InputRevisions, syn.InputRevisions) {
  return "InputsChanged", true
}
```

internal/controllers/synthesis/lifecycle.go
```diff
@@ -392,10 +394,43 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
...
+
+	reason := ""
+	// First condition
+	if isSynNil {
+		reason += "CurrentSynthesisEmpty"
+	} else {
+		if isCompGenDiff {
+			reason += "CompositionGenerationDiff"
+		} else if !isInputRevisionsEqual && isSynthesizedNotNil && !shouldIgnoreSideEffects {
+			reason += "InputRevisionsNotEqual && SynthesizedNotEmpty && ShouldNotIgnoreSideEffects"
```
jveski, line 415:
```suggestion
			reason += "InputsChanged"
```

internal/controllers/synthesis/lifecycle.go
```diff
@@ -392,10 +394,43 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
...
+	isCompInputsExist := comp.InputsExist(synth)
+	isCompInputsOutOfLockstep := comp.InputsOutOfLockstep(synth)
+
+	reason := ""
+	// First condition
+	if isSynNil {
+		reason += "CurrentSynthesisEmpty"
+	} else {
+		if isCompGenDiff {
+			reason += "CompositionGenerationDiff"
```
jveski, line 413:
```suggestion
			reason += "CompositionChanged"
```

internal/controllers/synthesis/lifecycle.go
```diff
@@ -392,10 +394,43 @@ func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
...
+	}
+	if !(isSynNil || isCompGenDiff || (!isInputRevisionsEqual && isSynthesizedNotNil && !shouldIgnoreSideEffects)) {
+		return "", false
+	}
+
+	// Second condition
+	if isCompDeleted {
+		reason += ", CompositionDeleted"
+	} else if isCompInputsExist && !isCompInputsOutOfLockstep {
+		reason += ", CompositionInputsExist && InputsNotOutOfLockstep"
```
jveski, line 426:
```suggestion
		reason += " && InputsInLockstep"
```

jveski (commented):
Leaving a few comments re. naming of the various conditions. I _think_ the logic looks right, would like a +1 from @AYM1607

internal/controllers/synthesis/lifecycle.go
```diff
@@ -381,7 +383,7 @@ func SwapStates(comp *apiv1.Composition) {
 	}
 }
 
-func shouldSwapStates(synth *apiv1.Synthesizer, comp *apiv1.Composition) bool {
+func shouldSwapStates(logger logr.Logger, synth *apiv1.Synthesizer, comp *apiv1.Composition) (string, bool) {
```
AYM1607, line 386: Logger is no longer needed

internal/controllers/synthesis/lifecycle.go
chihshenghuang: done

internal/controllers/synthesis/lifecycle.go
chihshenghuang: done

internal/controllers/synthesis/lifecycle.go
chihshenghuang: done

internal/controllers/synthesis/lifecycle_test.go
```diff
@@ -819,11 +819,19 @@ func TestShouldSwapStates(t *testing.T) {
 			},
 		},
 	}
+
 	for _, tc := range tests {
 		t.Run(tc.Name, func(t *testing.T) {
 			syn := &apiv1.Synthesizer{}
 			syn.Spec.Refs = []apiv1.Ref{{Key: "foo"}}
-			assert.Equal(t, tc.Expectation, shouldSwapStates(syn, &tc.Composition))
+			reason, shouldSwap := shouldSwapStates(syn, &tc.Composition)
```
AYM1607, line 827: Should we add the expected reason to the test cases?

chihshenghuang: change to verify the reason

AYM1607 (approved):
LGTM

jveski (approved):

PR #245, opened 2024-11-14
Readiness groups fixes

AYM1607:
- Dependencies were not correctly fetched for the default readiness group now that we support integers.
- Requeue resources that are waiting for the readiness of previous groups.


internal/controllers/reconciliation/controller.go
```diff
@@ -189,7 +189,7 @@ func (c *Controller) Reconcile(ctx context.Context, req *reconstitution.Request)
 			status := dep.FindStatus(slice)
 			if status == nil || status.Ready == nil {
 				logger.V(1).Info("skipping because at least one resource in an earlier readiness group isn't ready yet")
-				return ctrl.Result{}, nil
+				return ctrl.Result{RequeueAfter: wait.Jitter(c.readinessPollInterval, 0.1)}, nil
```
AYM1607, line 192: Wondering if we should allow this to be independently tuneable...

AYM1607:
This was not entirely the right approach, closing in favor of #248

PR #246, opened 2024-11-14
Hack to allow PDBs to progress

AYM1607:
This is needed because the label selector field on the [PDB spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#poddisruptionbudgetspec-v1-policy) has a patch strategy of "replace" meaning that the computed patch will never be empty even if the objects are the same, meaning that we'll never mark the resource as reconciled.


AYM1607:
Leaving this comment to block pending discussion on strategic-merge-diff @jveski

chihshenghuang (commented):
🚢

chihshenghuang (approved):

jveski (approved):

PR #247, opened 2024-11-15
Delete the composition when synthesizer is missing

### LINKED ISSUE #227: Composition deletion should succeed when its synthesizer is missing
There is no reason (beyond ease of implementation) to require the synthesizer in the deletion path.

chihshenghuang:
Resolves #227 

Delete the composition when synthesizer is missing


jveski (approved):
Nice fix! 💯

PR #248, opened 2024-11-16
Fix readiness groups lookup

AYM1607:
We used to represent readiness groups as unsigned integers, and the lookup logic had an optimization for the descending range starting at 0. This led always to always (incorrectly) returning no resources when looking up the previous readiness group before the default one.

We had another optimization based on tree node adjacency. However, we can't rely on it because this is a perfectly valid red-black tree:

image

Where the adjacent nodes are not necessarily parent + 1 and parent - 1 even if those nodes exist.


jveski (approved):
🥇 10/10 bug - great find!

PR #249, opened 2024-11-18
Johnaz/issue 239

johnazariah:
Added a gauge metric to keep track of the number of compositions waiting for inputs.
Uncovered the need for another gauge metric to keep track of the number of compositions without synthesizers.

### DIFF
--- a/internal/controllers/watchdog/controller.go
+++ b/internal/controllers/watchdog/controller.go
@@ -35,11 +35,23 @@ func (c *watchdogController) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, err
 	}
 
+	var inputsMissing int
+	var notInLockstep int
+	var withoutSynthesizers int
 	var pendingInit int
 	var pending int
 	var unready int
 	var terminal int
 	for _, comp := range list.Items {
+		if c.hasNoSynthesizer(&comp, ctx) {
+			withoutSynthesizers++
+		}
+		if c.waitingOnInputs(&comp, ctx) {
+			inputsMissing++
+		}
+		if c.getNotInLockstep(&comp, ctx) {
+			notInLockstep++
+		}
 		if c.pendingInitialReconciliation(&comp) {
 			pendingInit++
 		}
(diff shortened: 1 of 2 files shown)


internal/controllers/watchdog/controller.go
```diff
@@ -17,6 +18,19 @@ type watchdogController struct {
 	threshold time.Duration
 }
 
+func (c *watchdogController) getInputsExist(comp *apiv1.Composition, ctx context.Context) bool {
+	logger := logr.FromContextOrDiscard(ctx).WithValues("checkForMissingInputs: synthesizer access", comp.Generation)
```
jveski, line 22: This log column would be pretty noisy - remove it?

johnazariah: So how would we indicate that the thing returned true only because the synthesizer wasn't found?

internal/controllers/watchdog/controller.go
```diff
@@ -62,6 +77,32 @@ func (c *watchdogController) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+}
+
+func (c *watchdogController) hasNoSynthesizer(comp *apiv1.Composition, ctx context.Context) bool {
+	_, err := c.getSynthesizer(comp, ctx)
+	return err != nil
+}
+
+func (c *watchdogController) getInputsExist(comp *apiv1.Composition, ctx context.Context) bool {
+	syn, err := c.getSynthesizer(comp, ctx)
+	return (err != nil) || comp.InputsExist(syn)
```
jveski, line 94: Should this be `err == nil && comp.InputsExist(syn)` to match the logic of getNotInLockstep?

jveski:
LGTM pending one comment!

internal/controllers/watchdog/controller.go
johnazariah: No because if `err != nil` then `syn` is not available and from our discussion I got that we treat that condition as if inputs exist, not as if inputs don't exist...

jveski (approved):

PR #252, opened 2024-11-20
Remove reconciliation resource version cache

AYM1607:
Simplify this code to make it easier to reason about, also fixes a bug where resources were being marked as deleted in the resourceslice status on every resource version cache hit.


jveski (approved):
💯

PR #264, opened 2025-01-08
Fix deletion race

jveski:
It's possible for a resource to be deleted by an outside process between the time Eno gets the current state and attempts to delete it. In this case the client's delete func will return a 404 error, which we ignore. Since all errors from that function return `modified=false`, the controller doesn't requeue, so unless the resource has a reconciliation interval its status won't converge until the next Eno reconciler restart.


AYM1607 (approved):
😿

PR #272, opened 2025-01-22
Switch to structured merge diff

jveski:
The new structured merge diff libraries used by apiserver server-side apply are a nicer solution to strategic merge patch than the old approach. They expose the diff and allow callers to modify it before applying, which will enable partial reconciliation in the future (a very important feature for Eno).

Adopting structured merge diff also has the benefit of removing the `mungeDiff` function which has caused us much grief historically. It will immediately solve one known bug: removing properties doesn't currently work correctly.


internal/controllers/reconciliation/controller.go
```diff
@@ -266,81 +261,48 @@ func (c *Controller) reconcileResource(ctx context.Context, comp *apiv1.Composit
...
-	pdbGVK := schema.GroupVersionKind{Group: "policy", Version: "v1", Kind: "PodDisruptionBudget"}
-	if model == nil || (next != nil && next.GVK == pdbGVK) {
-		patch, err := jsonmergepatch.CreateThreeWayJSONMergePatch(prevJS, nextJS, currentJS)
-		if err != nil {
-			return nil, "", reconcile.TerminalError(err)
-		}
-		return patch, types.MergePatchType, err
+	if insecureLogPatch {
+		js, _ := updated.MarshalJSON()
+		logger.V(1).Info("INSECURE logging patch", "update", string(js))
```
jveski, line 295: I'll follow up with a PR to improve the logging here. Right now it just logs the entire resource.

internal/controllers/reconciliation/crud_test.go
```diff
@@ -861,7 +861,9 @@ func TestResourceDefaulting(t *testing.T) {
 									"name":  "foo",
 									"image": "bar",
 									"resources": map[string]any{
-										"memory": "1024Mi", // apiserver will return this as "1Gi"
+										"limits": map[string]any{
```
jveski, line 864: I'm not sure that this test ever worked correctly 😢

internal/discovery/cache.go
```diff
@@ -98,3 +122,45 @@ func (*Cache) evalVersion(ctx context.Context, v string) bool {
 
 	return major == 1 && minor >= 15
 }
+
+func parseGroupVersionKind(s proto.Schema) []schema.GroupVersionKind {
```
jveski, line 126: Existing code, moved from kubectl.go

internal/resource/resource.go
```diff
@@ -145,6 +133,115 @@ func (r *Resource) patchSetsDeletionTimestamp() bool {
 	return dt != ""
 }
 
+type SchemaGetter interface {
+	Get(ctx context.Context, gvk schema.GroupVersionKind) (typeref *smdschema.TypeRef, schem *smdschema.Schema, err error)
```
jveski, line 137: We pass a getter to Merge instead of just passing the values because it may need to invalidate the cache in the future in cases where the schema might be stale.

internal/resource/resource_test.go
```diff
@@ -213,3 +208,174 @@ func TestNewResource(t *testing.T) {
 		})
 	}
 }
+
+func TestMergeBasics(t *testing.T) {
```
jveski, line 212: The reconciliation controller has tests to cover most of this, but we should also expand test coverage here over time.

internal/controllers/reconciliation/controller.go
```diff
@@ -266,81 +261,48 @@ func (c *Controller) reconcileResource(ctx context.Context, comp *apiv1.Composit
...
-	patchmeta := strategicpatch.NewPatchMetaFromOpenAPI(model)
-	patch, err := strategicpatch.CreateThreeWayMergePatch(prevJS, nextJS, currentJS, patchmeta, true)
+	err = c.upstreamClient.Update(ctx, updated)
 	if err != nil {
-		return nil, "", reconcile.TerminalError(err)
+		return false, fmt.Errorf("applying update: %w", err)
 	}
-	return patch, types.StrategicMergePatchType, err
+
+	reconciliationActions.WithLabelValues("patch").Inc()
```
AYM1607, line 303: "update"?

internal/controllers/reconciliation/crud_test.go
AYM1607, line 864: Huh, memory is not even a valid field of the resources key right? How did it ever pass...

internal/controllers/reconciliation/edgecase_test.go
```diff
@@ -215,3 +218,113 @@ func TestLargeNamespaceDeletion(t *testing.T) {
...
+		return err == nil && comp.Status.CurrentSynthesis != nil && comp.Status.CurrentSynthesis.Ready != nil && comp.Status.CurrentSynthesis.ObservedSynthesizerGeneration == syn.Generation
+	})
+
+	// Prove the resource was reconciled correctly
+	cm := &corev1.ConfigMap{}
+	cm.Name = "test-obj"
+	cm.Namespace = "default"
+	err = mgr.DownstreamClient.Get(ctx, client.ObjectKeyFromObject(cm), cm)
+	require.NoError(t, err)
+	assert.Equal(t, map[string]string{"foo": "bar"}, cm.Labels)
```
AYM1607, line 328: Assert these before synth update just for completeness?

internal/controllers/reconciliation/edgecase_test.go
```diff
@@ -215,3 +218,113 @@ func TestLargeNamespaceDeletion(t *testing.T) {
 		return errors.IsNotFound(err)
 	}, time.Minute*3, time.Second)
 }
+
+// TestPatchStrategyReplace proves that resources which define a patch strategy of "replace" will eventually converge.
+func TestPatchStrategyReplace(t *testing.T) {
```
AYM1607, line 223: Add a comment to specify that PDBs in particular have this issue? I forgot for a second and was at a loss as to what was being tested.

internal/controllers/reconciliation/crud_test.go
jveski: Eno allows unknown properties so that should explain why it passed, I'm just not sure it could have failed lol

internal/controllers/reconciliation/controller.go
jveski: I'm hesitant to take a breaking change here without thinking a bit more about diff reporting first. Mind if I follow up on this when I improve patch/diff logging?

internal/discovery/cache.go
```diff
@@ -47,39 +51,59 @@ func (c *Cache) Get(ctx context.Context, gvk schema.GroupVersionKind) (proto.Sch
...
-				return nil, err
+				return nil, nil, err
 			}
 		}
 
-		model, ok := c.current[gvk]
+		model, ok := c.typeNameMap[gvk]
 		if !ok && c.fillWhenNotFound {
 			c.current = nil // invalidate cache - retrieve fresh schema on next attempt
-			discoveryCacheChanges.Inc()
```
AYM1607, line 57: We don't want to track this anymore?

jveski: I removed the metric since it doesn't seem that useful? We already log the same signal

internal/resource/resource.go
```diff
@@ -145,6 +133,115 @@ func (r *Resource) patchSetsDeletionTimestamp() bool {
...
+		toOld, err := typedOld.Compare(typedNew)
+		if err != nil {
+			return nil, false, fmt.Errorf("comparing new and old states: %w", err)
+		}
+		merged = merged.RemoveItems(toOld.Removed)
+	}
+
+	// Bail out if no changes are required
+	cmp, err := merged.Compare(typedCurrent)
+	if err == nil && cmp.IsSame() {
```
AYM1607, line 213: Should be bail early if err != nil?

jveski: I'm more inclined to fail open if we're for some reason unable to compare since an update applied multiple times is better than no update (in most cases). I'd like to think this code isn't reachable anyway but we all know how that goes

internal/controllers/reconciliation/controller.go
AYM1607: That's good on my end :)

AYM1607 (approved):

PR #273, opened 2025-01-24
Write buffer overhaul

jveski:
- Initialize the resource slice status using the typical patch instead of a separate/ad-hoc update to avoid meaningless conflicts/retries
- Protect strict ordering against zombie leaders using jsonpatch `test` ops
- Simplify batch timing - just use exponential backoff for back-to-back operations. The current approach is unnecessarily complex.
- Add fuzz/chaos style tests to build confidence in the concurrency safety of this implementation
- Keep track of how long items spend waiting on the buffer


cmd/eno-reconciler/main.go
```diff
@@ -48,7 +47,6 @@ func run() error {
 			DiscoveryRPS: 2,
 		}
 	)
-	flag.DurationVar(&writeBatchInterval, "write-batch-interval", time.Second*5, "The max throughput of composition status updates")
```
AYM1607, line 51: Mental note to remove this in AKS code during the next release.

internal/flowcontrol/writebuffer.go
```diff
@@ -34,29 +32,32 @@ type ResourceSliceWriteBuffer struct {
...
+func NewResourceSliceWriteBuffer(cli client.Client) *ResourceSliceWriteBuffer {
 	return &ResourceSliceWriteBuffer{
-		client: cli,
-		state:  make(map[types.NamespacedName][]*resourceSliceStatusUpdate),
+		client:        cli,
+		state:         make(map[types.NamespacedName][]*resourceSliceStatusUpdate),
+		insertionTime: make(map[types.NamespacedName]time.Time),
 		queue: workqueue.NewRateLimitingQueueWithConfig(
-			newRateLimiter(batchInterval, burst),
+			workqueue.NewItemExponentialFailureRateLimiter(time.Millisecond*100, 8*time.Second),
```
AYM1607, line 53: Since like this is deprecated in favor of https://pkg.go.dev/k8s.io/client-go/util/workqueue#NewTypedItemExponentialFailureRateLimiter in any case, are the base and max delays just heuristics or based on something?

internal/flowcontrol/writebuffer.go
```diff
@@ -65,16 +66,20 @@ func (w *ResourceSliceWriteBuffer) PatchStatusAsync(ctx context.Context, ref *re
...
 		}
 	}
 
+	w.insertionTime[key] = time.Now()
 	w.state[key] = append(currentSlice, &resourceSliceStatusUpdate{
 		SlicedResource: ref,
 		PatchFn:        patchFn,
 	})
-	w.queue.Add(key)
+	if w.queue.NumRequeues(key) == 0 {
```
AYM1607, line 80: Could you explain the reasoning behind this change? How does the queue behave when calling Add on an item that's already in it?

internal/flowcontrol/writebuffer.go
```diff
@@ -99,31 +104,43 @@ func (w *ResourceSliceWriteBuffer) processQueueItem(ctx context.Context) bool {
 	ctx = logr.NewContext(ctx, logger)
 
 	w.mut.Lock()
+	insertionTime := w.insertionTime[sliceNSN]
 	updates := w.state[sliceNSN]
 	delete(w.state, sliceNSN)
+
+	// Limit the number of operations per patch request
```
AYM1607, line 111: Where is the limit coming from?

internal/flowcontrol/writebuffer.go
```diff
@@ -99,31 +104,43 @@ func (w *ResourceSliceWriteBuffer) processQueueItem(ctx context.Context) bool {
...
+	if len(updates) > max {
+		w.state[sliceNSN] = updates[max:]
+		updates = updates[:max]
+	} else {
+		delete(w.insertionTime, sliceNSN)
+	}
 	w.mut.Unlock()
 
+	// We only forget the rate limit once the update queue for this slice is empty.
+	// So the first write is fast, but a steady stream of writes will be throttled exponentially.
```
AYM1607, line 122: For this to be true, shouldn't we always call AddRateLimited in PatchStatusAsync?

internal/flowcontrol/writebuffer.go
```diff
@@ -198,54 +172,70 @@ func (w *ResourceSliceWriteBuffer) updateSlice(ctx context.Context, sliceNSN typ
...
-}
-
-func (r *rateLimiter) When(item any) time.Duration {
-	r.failuresLock.Lock()
-	defer r.failuresLock.Unlock()
 
-	failures := r.failures[item]
-	r.failures[item]++
+	for _, update := range updates {
+		if update.SlicedResource.Index > len(slice.Spec.Resources)-1 || update.SlicedResource.Index < 0 {
```
AYM1607, line 204: Any reason to have it if it's impossible? And if it's just highly improbable, should we log?

internal/flowcontrol/writebuffer.go
```diff
@@ -198,54 +172,70 @@ func (w *ResourceSliceWriteBuffer) updateSlice(ctx context.Context, sliceNSN typ
...
-func (r *rateLimiter) NumRequeues(item interface{}) int {
-	r.failuresLock.Lock()
-	defer r.failuresLock.Unlock()
-	return r.failures[item]
+		path := fmt.Sprintf("/status/resources/%d", update.SlicedResource.Index)
+		patches = append(patches,
+			&jsonPatch{
+				Op:    "test", // make sure the current state is equal to the state we built the patch against
+				Path:  path,
+				Value: unsafeStatusPtr,
```
AYM1607, line 225: Very nice that this serializes as expected!

internal/flowcontrol/writebuffer.go
jveski: We're on an older version of controller-runtime that doesn't have generics yet 😢 . Created #274

jveski: Oh I just pulled these values out of the air while playing with some simple load tests. I don't think the timing is particularly important but maybe we should expose them as flags just in case?

internal/flowcontrol/writebuffer.go
jveski: Calling `Add` while the same key is in-flight will cause it to be requeued after `Done` is called regardless of the rate limiter. `AddRateLimited` ultimately just calls `Add` but increments honors the rate limiter state.

In this case the first attempt should count against the rate limit and calls to PatchStatusAsync shouldn't extend or reset the rate limit. Then we increment the retry counter (in `processQueueItem`) before calling Done to back off any hot loops.

internal/flowcontrol/writebuffer.go
jveski: This beauty: https://github.com/kubernetes/apiserver/blob/51262c6edb897d1063d2b02183895a10f7132b8f/pkg/endpoints/handlers/patch.go#L60

I thought the request body size would be too large anyway but the integration test proves otherwise 🤷 . It works perfectly right at the limit.

internal/flowcontrol/writebuffer.go
jveski: That would back off once per call to `PatchStatusAsync` which seems too aggressive. Right now we only back off once per batch i.e. call to `updateSlice`.

internal/flowcontrol/writebuffer.go
```diff
@@ -99,31 +104,43 @@ func (w *ResourceSliceWriteBuffer) processQueueItem(ctx context.Context) bool {
...
 	w.state[sliceNSN] = append(updates, w.state[sliceNSN]...)
+	w.insertionTime[sliceNSN] = insertionTime
 	w.mut.Unlock()
 	w.queue.AddRateLimited(item)
 
 	return true
 }
 
-func (w *ResourceSliceWriteBuffer) updateSlice(ctx context.Context, sliceNSN types.NamespacedName, updates []*resourceSliceStatusUpdate) bool {
+func (w *ResourceSliceWriteBuffer) updateSlice(ctx context.Context, insertionTime time.Time, sliceNSN types.NamespacedName, updates []*resourceSliceStatusUpdate) bool {
```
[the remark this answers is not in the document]
jveski: Strangely enough I have few opinions about named returns - will add one 👍

internal/flowcontrol/writebuffer.go
jveski: If this somehow did evaluate to true in production, I worry that it would generate a ton of logs. I'm personally okay with swallowing this condition but open to other options

internal/flowcontrol/writebuffer.go
AYM1607: I'll leave it up to you, would be nice to have the ability to overwrite them in prod, but I wouldn't mind leaving it to a follow up.

internal/flowcontrol/writebuffer.go
AYM1607: Lol, wonder whey they don't export that var...

internal/flowcontrol/writebuffer.go
AYM1607: Swallowing is fine for now, though I wonder if the fact that it was true in prod reflects on a deeper issue we're not seeing

AYM1607 (approved):

PR #276, opened 2025-01-31
Improve the synthesis dispatcher implementation

jveski:
Merge the two rollout controllers and concurrency limit controller into a new "scheduling" controller responsible for making all synthesis dispatch decisions. The synthesis lifecycle controller is finally free to focus on pod lifecycle.

The focus here is on fuzz testing to cover edge cases and concurrency bugs. A ton of small papercut bugs have been fixed - not worth enumerating here.


internal/controllers/aggregation/composition.go
```diff
@@ -72,11 +72,6 @@ func (c *compositionController) aggregate(synth *apiv1.Synthesizer, comp *apiv1.
 		return copy
 	}
 
-	if comp.Status.CurrentSynthesis.UUID == "" {
```
jveski, line 75: This state doesn't exist anymore - the UUID is now set when swapping the syntheses since there's no need for the second dispatch phase

internal/controllers/aggregation/composition.go
```diff
@@ -108,9 +103,6 @@ func (c *compositionController) aggregate(synth *apiv1.Synthesizer, comp *apiv1.
 	if comp.Status.CurrentSynthesis.Ready != nil {
 		copy.Status = "Ready"
 	}
-	if comp.Status.PendingResynthesis != nil {
-		copy.Status = "WaitingForCooldown"
```
jveski, line 112: This was too chatty anyway - generated tons of writes just to update a relatively short-lived and unimportant status message. In any case, `PendingResynthesis ` is going away ⚡

internal/controllers/scheduling/op.go
```diff
@@ -0,0 +1,233 @@
...
+	Composition *apiv1.Composition
+	Reason      opReason
+
+	Dispatched time.Time
+
+	id               uuid.UUID // set when patch is built
+	synthRolloutHash []byte    // memoized
+}
+
+func newOp(synth *apiv1.Synthesizer, comp *apiv1.Composition) *op {
```
jveski, line 29: This implements the logic of the old `shouldSwapStates` function - just (IMO) much easier to grok.

I promise it passes the old unit tests. They were wired up during refactoring but I removed them since the new tests provide better coverage anyway.

api/v1/composition.go
```diff
@@ -214,3 +214,32 @@ func (s *CompositionStatus) GetCurrentSynthesisUUID() string {
...
+func (c *Composition) EnableIgnoreSideEffects() {
+	anno := c.GetAnnotations()
+	if anno == nil {
+		anno = map[string]string{}
+	}
+	anno["eno.azure.io/ignore-side-effects"] = "true"
+	c.SetAnnotations(anno)
+}
+
+const forceResynthesisAnnotation = "eno.azure.io/force-resynthesis"
```
AYM1607, line 231: I'm pretty sure I'll get an answer after reading the rest of the code, but I'll leave this here in case the meaning is not clear later on.

api/v1/composition.go
```diff
@@ -214,3 +214,32 @@ func (s *CompositionStatus) GetCurrentSynthesisUUID() string {
...
+
+func (c *Composition) ForceResynthesis() {
```
AYM1607, line 233: Does this mean that we can only fore resynthesis once?

internal/controllers/scheduling/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	op.Dispatched = time.Now()
+	c.lastApplied = op
+	logger.V(0).Info("dispatched synthesis")
+
+	return ctrl.Result{}, nil
+}
+
+// getNextCooldownSlot returns the next time at which a deferred synthesis can be dispatched while honoring the configured cooldown period.
+func (c *controller) getNextCooldownSlot(comps *apiv1.CompositionList) time.Time {
+	var last time.Time
```
AYM1607, line 156: Just FYI that I had to read this a few times to grok it, I was thining about `last` as the one with the oldest initialized. In my head `newest` would make more sense, but with this kind of things it gets subjective real fast...

internal/controllers/scheduling/op.go
```diff
@@ -0,0 +1,233 @@
...
+
+type op struct {
+	Synthesizer *apiv1.Synthesizer
+	Composition *apiv1.Composition
+	Reason      opReason
+
+	Dispatched time.Time
+
+	id               uuid.UUID // set when patch is built
+	synthRolloutHash []byte    // memoized
```
AYM1607, line 26: Any particular concurrency reason to memoize? Or is it just to avoid the hash computation?

internal/controllers/scheduling/op.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func newOp(synth *apiv1.Synthesizer, comp *apiv1.Composition) *op {
+	syn := comp.Status.CurrentSynthesis
+	o := &op{Synthesizer: synth, Composition: comp}
+
+	switch {
+	case comp.DeletionTimestamp != nil || !comp.InputsExist(synth) || comp.InputsOutOfLockstep(synth) || !controllerutil.ContainsFinalizer(comp, "eno.azure.io/cleanup"):
```
AYM1607, line 34: A few questions on this one:
- Is the finalizer guard here to protect against compositions that haven't been seen by Eno yet or ones whose finalizer has been just removed? If it's the former (which is my guess) what goes wrong when we don't guard it?
- Is our reliance on the input controller and the fact that this could get stuck the reason for the op timeout in the controller?

internal/controllers/scheduling/op.go
```diff
@@ -0,0 +1,233 @@
...
+
+	case syn.ObservedCompositionGeneration != comp.Generation:
+		o.Reason = compositionModifiedOp
+		return o
+
+	case comp.ShouldIgnoreSideEffects():
+		return nil
+	}
+
+	nonDeferredInputChanges, deferredInputChanges := inputChangeCount(synth, comp.Status.InputRevisions, syn.InputRevisions)
```
AYM1607, line 56: Any particular reason these are outside the switch? I assume that it's beacuse it relies on syn not being nil?

internal/controllers/scheduling/op.go
```diff
@@ -0,0 +1,233 @@
...
+	ops := []jsonPatch{}
+
+	if o.id == uuid.Nil {
+		// defer generating the uuid until we know the op is definitely going to be dispatched
+		o.id = uuid.Must(uuid.NewRandom())
+	}
+
+	if reflect.DeepEqual(o.Composition.Status, apiv1.CompositionStatus{}) {
+		ops = append(ops,
+			jsonPatch{Op: "test", Path: "/status", Value: nil},
```
AYM1607, line 128: Kinda cool that the individual patches get applied sequentially!

internal/controllers/scheduling/op.go
```diff
@@ -0,0 +1,233 @@
...
+			jsonPatch{Op: "add", Path: "/status", Value: map[string]any{}})
+	}
+
+	ops = append(ops, jsonPatch{Op: "test", Path: "/status/inputRevisions", Value: o.Composition.Status.InputRevisions})
```
AYM1607, line 132: Comment on why this is important? I guess it goes back to synchronization? The inputs controller can write in-between the patch build and when it actually gets applied

api/v1/composition.go
jveski: Currently we set the pending synthesis timestamp to force resynth if one of the slices is missing. Using this annotation allows us to target a specific synthesis. So if the states happened to be swapped concurrently we don't synthesize twice.

I also like the idea that it's an undocumented private API rather than being part of the status, but either way is probably fine here.

api/v1/composition.go
jveski: Only one synthesis UUID can be eligible for resynthesis at once, which I think makes sense?

internal/controllers/scheduling/controller.go
jveski: I think in this case you're just objectively correct 😆 . Fixed.

internal/controllers/scheduling/op.go
jveski: Just to avoid hashing. Because `SynthRolloutOrderHash` is called by `Less`, it will likely be called several times per op.

internal/controllers/scheduling/op.go
jveski: > Is the finalizer guard here to protect against compositions that haven't been seen by Eno yet or ones whose finalizer has been just removed? If it's the former (which is my guess) what goes wrong when we don't guard it?

I think this would technically be safe without guarding on the finalizer, since swapping the state doesn't necessarily mean there will be anything to clean up on deletion. The synthesis lifecycle controller will add the finalizer before creating the pod anyway.

For whatever reason removing this check caused one of the batch reconciliation tests to fail so here it stays 🤷 .

internal/controllers/scheduling/op.go
jveski: Yeah exactly. And the multiple return values make it awkward in the case statement anyway

internal/controllers/scheduling/op.go
```diff
@@ -0,0 +1,233 @@
...
+
+	return o.Reason < than.Reason
+}
+
+// SynthRolloutOrderHash returns a hash that represents this composition's order in the rollout of a particular synthesizer generation.
+// This mechanism maintains determinism while shuffling the rollout order of every synthesizer change.
+func (o *op) SynthRolloutOrderHash() []byte {
+	if o.synthRolloutHash == nil {
+		hash := fnv.New64()
+		fmt.Fprintf(hash, "%s:%d:%s", o.Synthesizer.UID, o.Synthesizer.Generation, o.Composition.UID)
```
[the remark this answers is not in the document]
jveski: The UID doesn't change when the synth changes so (unless I'm missing something lol) the rollout order would be the same every time if we didn't include the generation.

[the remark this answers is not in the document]
AYM1607: Ah sorry, I meant the other way around the UID is not necessary, but the Generation is, in any case, it doesn't hurt

api/v1/composition.go
AYM1607: I don't think I saw this code, but does the annotation get cleared when the current synthesis succeeds? I ask this mainly because the self healing controller relies on being able to force a resynthesis right? I kinda dislike the idea of the self-healing controller to be honest, it's just a manifestation of races/inconsistencies somewhere else. If we still need it though, I want to make sure that it's still supported

jveski: The annotation doesn't get cleared, it just becomes a no-op since once it's accepted the UUID will no longer match.

Spoiler alert on the change I'm working on next: the self healing controller gets merged into a new more generic "resource slice controller" along with a couple of others 😊

internal/controllers/scheduling/op.go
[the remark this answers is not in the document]
jveski: Ahh that make sense. I think we could get by without the UID, but it's harder to reason about what happens when a composition switches to another synthesizer (unlikely as that is). Including the synth UID simplifies the hash conceptually IMO - albeit at a slight computational cost.

api/v1/composition.go
AYM1607: That's perfect, lgtm then!

internal/controllers/scheduling/op.go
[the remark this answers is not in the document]
AYM1607: Good point about a composition potentially switching to another synth, I haven't thought through how that'd be implemented, this might or might not be involved in that process, but I think the cost is negligible, so I'm good or keeping it around.

AYM1607 (approved):

PR #281, opened 2025-02-07
Synthesis priority bugfix

jveski:
Fixes a priority sorting bug in the scheduling controller, slightly refactors for clarity, and improves fuzz test coverage for the `newOp` function.

### DIFF
--- a/internal/controllers/scheduling/controller.go
+++ b/internal/controllers/scheduling/controller.go
@@ -113,7 +113,7 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 
 		next := newOp(&synth, &comp)
-		if next != nil && (op == nil || op.Less(next)) {
+		if next != nil && (op == nil || next.Less(op)) {
 			op = next
 		}
 	}
@@ -138,15 +138,14 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	if err := c.dispatchOp(ctx, op); err != nil {
 		if errors.IsInvalid(err) {
-			logger.V(0).Info("conflict while dispatching synthesis")
-			return ctrl.Result{}, nil
+			return ctrl.Result{}, fmt.Errorf("conflict while dispatching synthesis")
 		}
 		return ctrl.Result{}, fmt.Errorf("dispatching synthesis operation: %w", err)
 	}
 
 	op.Dispatched = time.Now()
 	c.lastApplied = op
-	logger.V(0).Info("dispatched synthesis")
+	logger.V(0).Info("dispatched synthesis", "synthesisUUID", op.id)
 
 	return ctrl.Result{}, nil
 }
--- a/internal/controllers/scheduling/controller_test.go
+++ b/internal/controllers/scheduling/controller_test.go
@@ -485,19 +485,21 @@ func TestDispatchOrder(t *testing.T) {
 		Synthesized:                   ptr.To(metav1.Now()),
 	}
 
+	// comp1 is ready for resynthesis because its spec has changed since its last synthesis
 	comp1 := comp.DeepCopy()
 	comp1.Name = "test-comp-1"
 	comp1.Status.CurrentSynthesis.ObservedCompositionGeneration--
 	require.NoError(t, cli.Create(ctx, comp1))
 	require.NoError(t, cli.Status().Update(ctx, comp1))
 
+	// comp2 is ready for synthesis because its synthesizer has changed since its last synthesis
 	comp2 := comp.DeepCopy()
 	comp2.Name = "test-comp-2"
 	comp2.Status.CurrentSynthesis.ObservedSynthesizerGeneration--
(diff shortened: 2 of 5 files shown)


internal/controllers/scheduling/controller.go
```diff
@@ -138,15 +138,14 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	if err := c.dispatchOp(ctx, op); err != nil {
 		if errors.IsInvalid(err) {
-			logger.V(0).Info("conflict while dispatching synthesis")
-			return ctrl.Result{}, nil
+			return ctrl.Result{}, fmt.Errorf("conflict while dispatching synthesis")
```
jveski, line 141: It shouldn't be possible that we need to requeue on conflict, but if there is any kind of trivial bug in the `op` logic it becomes more likely. So it's safer to just requeue to mitigate the risk.

internal/controllers/scheduling/controller_test.go
```diff
@@ -469,32 +469,50 @@ func TestDispatchOrder(t *testing.T) {
...
+	comp1.Name = "test-comp-1"
+	comp1.Status.CurrentSynthesis.ObservedCompositionGeneration--
+	require.NoError(t, cli.Create(ctx, comp1))
+	require.NoError(t, cli.Status().Update(ctx, comp1))
 
 	comp2 := comp.DeepCopy()
 	comp2.Name = "test-comp-2"
-	comp2.Status.CurrentSynthesis.ObservedSynthesizerGeneration = synth.Generation
-	require.NoError(t, cli.Create(ctx, comp))
+	comp2.Status.CurrentSynthesis.ObservedSynthesizerGeneration--
```
jveski, line 496: The logic of this test was subtly wrong before which is how the sorting bug was able to merge.

internal/controllers/scheduling/op.go
```diff
@@ -27,48 +27,55 @@ type op struct {
 }
 
 func newOp(synth *apiv1.Synthesizer, comp *apiv1.Composition) *op {
-	syn := comp.Status.CurrentSynthesis
 	o := &op{Synthesizer: synth, Composition: comp}
 
+	var ok bool
+	o.Reason, ok = classifyOp(synth, comp, comp.Status.CurrentSynthesis)
```
jveski, line 33: Small refactor: move the reason logic to `classifyOp`.

internal/controllers/scheduling/op.go
```diff
@@ -27,48 +27,55 @@ type op struct {
...
 	}
 
 	nonDeferredInputChanges, deferredInputChanges := inputChangeCount(synth, comp.Status.InputRevisions, syn.InputRevisions)
-	if nonDeferredInputChanges > 0 && syn.Synthesized != nil {
-		o.Reason = inputModifiedOp
-		return o
+	if nonDeferredInputChanges > 0 {
+		return inputModifiedOp, true
 	}
-	if deferredInputChanges > 0 && syn.Synthesized != nil {
```
jveski, line 61: The `syn.Synthesized != nil` logic is handled upstream of this function now but the behavior is preserved.

internal/controllers/scheduling/op.go
```diff
@@ -27,48 +27,55 @@ type op struct {
...
 	case syn == nil:
-		o.Reason = initialSynthesisOp
-		return o
+		return initialSynthesisOp, true
 
 	case comp.ShouldForceResynthesis():
-		o.Reason = forcedResynthesisOp
-		return o
-
-	case comp.Synthesizing():
```
jveski, line 45: This was confusing so it's replaced with the ` o.Reason.Deferred() && comp.Synthesizing()` expression in `newOp`

internal/controllers/scheduling/controller.go
```diff
@@ -113,7 +113,7 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 
 		next := newOp(&synth, &comp)
-		if next != nil && (op == nil || op.Less(next)) {
+		if next != nil && (op == nil || next.Less(op)) {
```
jveski, line 116: lolololol

internal/controllers/scheduling/controller_test.go
```diff
@@ -469,32 +469,50 @@ func TestDispatchOrder(t *testing.T) {
...
 	comp.Spec.Synthesizer.Name = synth.Name
-	comp.Status.CurrentSynthesis = &apiv1.Synthesis{UUID: "foo", ObservedCompositionGeneration: 1, Synthesized: ptr.To(metav1.Now())}
+	comp.Status.CurrentSynthesis = &apiv1.Synthesis{
+		UUID:                          "foo",
+		ObservedCompositionGeneration: comp.Generation,
+		ObservedSynthesizerGeneration: synth.Generation,
+		Synthesized:                   ptr.To(metav1.Now()),
+	}
+
+	comp1 := comp.DeepCopy()
```
AYM1607, line 488: Care to add some comments to explain what each of these comps represent what what part of the controller they're testing?

AYM1607 (approved):

PR #282, opened 2025-02-13
Resource tree

jveski:
This is part of what will ultimately replace the reconstitution cache. The goal is to clarify and more carefully test how resources within a given synthesis are indexed.

Nothing is modified by this PR - the code it adds will be wired up later.


api/v1/resourceslice.go
```diff
@@ -44,6 +44,25 @@ type ResourceState struct {
 	Deleted    bool         `json:"deleted,omitempty"`
 }
 
+func (r *ResourceState) Equal(rr *ResourceState) bool {
```
AYM1607, line 47: Does DeepEqual not work here? Or do we care about these semantics for some concrete reason?

jveski: It would work, I'm just concerned about perf since this is called for every resource every time the slice changes

internal/resource/resource.go
```diff
@@ -38,6 +38,10 @@ type Ref struct {
 	Name, Namespace, Group, Kind string
 }
 
+func (r *Ref) String() string {
+	return fmt.Sprintf("(%s.%s)/%s/%s", r.Kind, r.Group, r.Namespace, r.Name)
```
AYM1607, line 42: nit: Grop.Kind?

internal/resource/resource.go
```diff
@@ -348,6 +352,15 @@ func NewResource(ctx context.Context, renv *readiness.Env, slice *apiv1.Resource
 	return res, nil
 }
 
+// Less returns true when r < than.
+// Used to establish determinstic ordering for conflicting resources.
+func (r *Resource) Less(than *Resource) bool {
+	if r.Manifest == nil || than.Manifest == nil {
+		return true
+	}
+	return r.Manifest.Manifest < than.Manifest.Manifest
```
AYM1607, line 361: Is the string representation deterministic?

internal/resource/tree.go
```diff
@@ -0,0 +1,164 @@
...
+	State               *apiv1.ResourceState
+	PendingDependencies map[Ref]struct{}
+	Dependents          map[Ref]*indexedResource
+}
+
+// treeBuilder is used to index a set of resources into a stateTree.
+type treeBuilder struct {
+	byRef        map[Ref]*indexedResource
+	byGroup      *redblacktree.Tree[int, []*indexedResource]
+	byDefiningGK map[schema.GroupKind]*indexedResource
```
AYM1607, line 23: byDefiningGK == CRs? maybe a comment as to why we neeed both?

internal/resource/tree.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if b.byDefiningGK == nil {
+		b.byDefiningGK = map[schema.GroupKind]*indexedResource{}
+	}
+	if b.byGK == nil {
+		b.byGK = map[schema.GroupKind]*indexedResource{}
+	}
+
+	// Handle conflicting refs deterministically
+	if existing, ok := b.byRef[resource.Ref]; ok && resource.Less(existing.Resource) {
```
AYM1607, line 43: Should we be referencing a field within the manifest rather than the whole manifest? Or does it not matter which one wins for the purposes of this tree?

internal/resource/tree.go
```diff
@@ -0,0 +1,164 @@
...
+
+	for _, idx := range b.byRef {
+		t.byManiRef[idx.Resource.ManifestRef] = idx
+
+		// CRs are dependent on their CRDs
+		i := b.byGroup.IteratorAt(b.byGroup.GetNode(idx.Resource.ReadinessGroup))
+		crd, ok := b.byDefiningGK[idx.Resource.GVK.GroupKind()]
+		if ok {
+			idx.PendingDependencies[crd.Resource.Ref] = struct{}{}
+			crd.Dependents[idx.Resource.Ref] = idx
```
AYM1607, line 76: Double link for efficiency?

internal/resource/tree.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		// Depend on any resources in the previous readiness group
+		if i.Prev() {
+			for _, dep := range i.Value() {
+				idx.PendingDependencies[dep.Resource.Ref] = struct{}{}
+			}
+		}
+		i.Next() //rewind
```
AYM1607, line 85: Only rewind if prev succeeded?

AYM1607: Or does it awlays go back even if there's nothing?

internal/resource/tree.go
```diff
@@ -0,0 +1,164 @@
...
+		// CRs are dependent on their CRDs
+		i := b.byGroup.IteratorAt(b.byGroup.GetNode(idx.Resource.ReadinessGroup))
+		crd, ok := b.byDefiningGK[idx.Resource.GVK.GroupKind()]
+		if ok {
+			idx.PendingDependencies[crd.Resource.Ref] = struct{}{}
+			crd.Dependents[idx.Resource.Ref] = idx
+		}
+
+		// Depend on any resources in the previous readiness group
+		if i.Prev() {
```
AYM1607, line 80: I have PTSD from the last bug, Prev and Next don't rely on adjacency right?

internal/resource/tree.go
jveski: The iterator is weird: it will progress regardless of if there is an item in that position. So Next is needed even if Prev returns false. Will add a comment.

internal/resource/tree.go
jveski: lolol Right - the iterator behaves correctly in this case. Just hopping between the nodes using their pointers is where we got into trouble before.

internal/resource/tree.go
jveski: It doesn't matter which wins as long as the same one will win every time regardless of order

internal/resource/tree.go
jveski: will do!

internal/resource/resource.go
jveski: Yes, since we only care about determinism within the context of a particular synthesis. So if the synthesizer moves around the order of some properties in the next synthesis it doesn't matter from this func's perspective.

AYM1607 (approved):

PR #284, opened 2025-02-18
helmshim: call write() for default output writer only in RenderChart func

chihshenghuang:
Call write() for default output writer only in RenderChart func. If the client provides an output writer, it should handle write() itself for flexibility.

### DIFF
--- a/pkg/helmshim/helm_test.go
+++ b/pkg/helmshim/helm_test.go
@@ -2,6 +2,7 @@ package helmshim
 
 import (
 	"bytes"
+	"os"
 	"testing"
 
 	"github.com/Azure/eno/pkg/function"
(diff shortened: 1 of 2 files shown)


tilnl (approved):

pkg/helmshim/helm_test.go
```diff
@@ -16,12 +17,38 @@ func TestRenderChart(t *testing.T) {
...
+	err = RenderChart(WithInputReader(i), WithChartPath("fixtures/basic-chart"))
+	require.NoError(t, err)
+
+	// Close the writer and capture the output from reader.
+	w.Close()
+	var buf bytes.Buffer
+	buf.ReadFrom(r)
+	output := buf.String()
+	// Restore the original stdout.
+	os.Stdout = oldStdout
```
jveski, line 46: We should defer this right after overriding it so the value will be reset even if the test fails

pkg/helmshim/helm_test.go
```diff
@@ -16,12 +17,38 @@ func TestRenderChart(t *testing.T) {
...
+	err = o.Write()
+	require.NoError(t, err)
 	assert.Equal(t, "{\"apiVersion\":\"config.kubernetes.io/v1\",\"kind\":\"ResourceList\",\"items\":[{\"apiVersion\":\"v1\",\"data\":{\"input\":\"{\\\"apiVersion\\\":\\\"v1\\\",\\\"kind\\\":\\\"ConfigMap\\\",\\\"metadata\\\":{\\\"annotations\\\":{\\\"eno.azure.io/input-key\\\":\\\"foo\\\"},\\\"name\\\":\\\"test-cm\\\"}}\",\"inputResourceName\":\"test-cm\",\"some\":\"value\"},\"kind\":\"ConfigMap\",\"metadata\":{\"name\":null}},{\"apiVersion\":\"somegroup.io/v9001\",\"kind\":\"ATypeNotKnownByTheScheme\",\"metadata\":{\"name\":\"foo\"}}]}\n", output.String())
 }
 
+func TestRenderChartWithDefaultOutputWriter(t *testing.T) {
+	// Save the original stdout and redirect it to a pipe.
+	oldStdout := os.Stdout
+	r, w, _ := os.Pipe()
+	os.Stdout = w
```
jveski, line 31: This is clever, but I'm wondering if it might have some strange interactions with the test runner, since it might write to stdout. I wouldn't block on this, but would slightly prefer to define a global var in this package like `var stdout = os.Stdout`, use it as the output writer default, and override only that value from the tests.

pkg/helmshim/helm_test.go
chihshenghuang: update! good call that the test fails will cause the stdout didn't get restore back

pkg/helmshim/helm_test.go
chihshenghuang: agree with the concern, i change to use package level stdout to create the output writer to simulate the behavior of default output writer which uses os.Stdout.

chihshenghuang: discuss offline, remove the test

jveski (approved):

PR #288, opened 2025-02-19
Simplify the resource cache

jveski:
This PR replaces the resource cache out from underneath the reconstitution controller without changing any fundamental architectural patterns, etc.

The new cache is simpler to reason about and has significantly better test coverage. The most significant change is that the status of each resource is now fed back from the informer to the reconstitution cache, which ends up simplifying the reconciliation controller.


cmd/eno-reconciler/main.go
```diff
@@ -108,15 +109,15 @@ func run() error {
...
 
-	rCache := reconstitution.NewCache(mgr.GetClient())
 	recOpts.Manager = mgr
-	recOpts.Cache = rCache
 	recOpts.WriteBuffer = writeBuffer
 	recOpts.Downstream = remoteConfig
 	recOpts.Queue = workqueue.NewTypedRateLimitingQueue(
-		workqueue.DefaultTypedItemBasedRateLimiter[reconstitution.Request]())
+		workqueue.DefaultTypedItemBasedRateLimiter[resource.Request]())
+	recOpts.Cache = &resource.Cache{}
```
AYM1607, line 117: Nit: Even if no options are being set now, maybe put it under a constructor? Just so it's harder to forget

internal/controllers/reconciliation/controller.go
```diff
@@ -115,58 +115,23 @@ func (c *Controller) Reconcile(ctx context.Context, req reconstitution.Request)
 	ctx = logr.NewContext(ctx, logger)
 
 	// Find the current and (optionally) previous desired states in the cache
-	synRef := reconstitution.NewSynthesisRef(comp)
-	resource, exists := c.resourceClient.Get(ctx, synRef, &req.Resource)
+	var prev *resource.Resource
+	resource, visible, exists := c.resourceClient.Get(ctx, comp.Status.GetCurrentSynthesisUUID(), req.Resource)
 	if !exists {
 		// It's possible for the cache to be empty because a manifest for this resource no longer exists at the requested composition generation.
 		// Dropping the work item is safe since filling the new version will generate a new queue message.
```
AYM1607, line 122: Unrelated to this PR, but what does "filling the new version" mean here?

internal/controllers/reconciliation/controller.go
```diff
@@ -195,23 +160,6 @@ func (c *Controller) Reconcile(ctx context.Context, req reconstitution.Request)
 		ready = status.Ready
 	}
 
-	// Evaluate the readiness of resources in the previous readiness group
```
AYM1607, line 198: this is all enclosed in op now?

internal/resource/resource.go
```diff
@@ -336,7 +336,7 @@ func NewResource(ctx context.Context, renv *readiness.Env, slice *apiv1.Resource
 			name = "default"
 		}
 
-		check, err := readiness.ParseCheck(renv, value)
+		check, err := readiness.ParseCheck(readiness.DefaultEnv, value)
```
AYM1607, line 339: Not worth the injection?

internal/reconstitution/controller.go
```diff
@@ -64,46 +54,58 @@ func (r *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	ctx = logr.NewContext(ctx, logger)
 
-	// We populate the cache with both the previous and current syntheses
-	prevReqs, err := r.populateCache(ctx, comp, comp.Status.PreviousSynthesis)
+	filled, err := r.populateCache(ctx, comp, comp.Status.PreviousSynthesis)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("processing previous state: %w", err)
 	}
-	currentReqs, err := r.populateCache(ctx, comp, comp.Status.CurrentSynthesis)
+	if filled {
```
AYM1607, line 61: This reads a bit weird, does filled mean that it was filled on this tick? Or that it was already filled when observed on this tick?

internal/controllers/reconciliation/controller.go
jveski: Yeahh that comment is kind of unclear. IIRC the idea is that it's safe to drop the work item because filling the cache will always generate another work item anyway

internal/resource/resource.go
jveski: hmm maybe we should just use the default env from ParseCheck and make it private. Will investigate.

internal/controllers/reconciliation/controller.go
jveski: It's all handled by the synthesis cache and resource tree concepts so the reconciliation controller doesn't need to worry about the dependency graph.

internal/reconstitution/controller.go
jveski: will reply in the other related comment

internal/reconstitution/controller.go
```diff
@@ -64,46 +54,58 @@ func (r *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		}
+		slices[i] = slice
 	}
 
 	logger = logger.WithValues("synthesisCompositionGeneration", synthesis.ObservedCompositionGeneration)
+	compNSN := types.NamespacedName{Namespace: comp.Namespace, Name: comp.Name}
 	ctx = logr.NewContext(ctx, logger)
-	if r.Cache.hasSynthesis(comp, synthesis) {
-		return nil, nil
+	if r.cache.Visit(ctx, comp, synthesis.UUID, slices) {
```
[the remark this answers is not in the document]
jveski: I'd love for it to be more obvious in the type signatures but in any case here's the logic:

- Visit returns true _only_ when the cache needs to be filled
- populateCache returns true when the cache was filled during that tick of the loop

So Visit populates the status of the resources in the tree, Fill populates the spec. And Fill takes the full resources from the api client, Visit takes the trimmed resources from the informers. That's why we need to requeue after filling the cache-  to get the status.

All of this is to the end of only reading status from the informers - otherwise it's possible for the status to backtrack (apply newer status from the api client, then from stale informer). But in any case lmk if you can think of any way to make the behavior more obvious.

internal/resource/resource.go
jveski: Refactored some things to encap the cel env entirely within the readiness package

internal/controllers/reconciliation/controller.go
jveski: Updated the comment for clarity

internal/reconstitution/controller.go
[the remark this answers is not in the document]
AYM1607: > Visit returns true only when the cache needs to be filled

Wouldn't then we need to proceed instead of returning early here?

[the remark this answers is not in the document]
jveski: Ahh I had it backwards in my description, here's the comment:

```go
// Visit takes a set of resource slices from the informers and updates the resource status in the cache.
// Return false if the synthesis is not in the cache.
func (c *Cache) Visit(ctx context.Context, comp *apiv1.Composition, synUUID string, items []apiv1.ResourceSlice) bool {
```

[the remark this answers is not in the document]
jveski: So code is right, I'm wrong 😢

AYM1607 (approved):

PR #289, opened 2025-02-21
Remove bottleneck in reconciliation controller

jveski:
It turns out the default controller work queue rate limiter has a shared token bucket of 10 RPS (with some reasonable burst) that applies to all requeues. Since the reconciliation controller can't/doesn't watch the resources it writes, it relies on requeues to form its feedback loop, which means when the loop ticks into the second iteration of any resource (typically the status update) it will wait a few dozen milliseconds.

This also fixes a couple of other small nits surrounding client-side rate limiting.


cmd/eno-reconciler/main.go
```diff
@@ -95,14 +96,14 @@ func run() error {
...
+	remoteConfig := rest.CopyConfig(mgr.GetConfig())
 	if remoteKubeconfigFile != "" {
 		if remoteConfig, err = k8s.GetRESTConfig(remoteKubeconfigFile); err != nil {
 			return err
 		}
-		if remoteQPS != 0 {
-			remoteConfig.QPS = float32(remoteQPS)
-		}
+	}
+	if remoteQPS >= 0 {
```
jveski, line 105: The configured QPS should apply to the downstream client regardless of if its config was inferred from the environment or specified explicitly

internal/controllers/reconciliation/helpers_test.go
```diff
@@ -52,3 +58,42 @@ func writeComposition(t *testing.T, client client.Client, orphan bool) (*apiv1.S
...
+		Timeout:               time.Minute,
+		ReadinessPollInterval: time.Hour,
+	})
+	require.NoError(t, err)
+
+	err = reconstitution.New(mgr.Manager, &cache, queue)
+	require.NoError(t, err)
+}
+
+func mapToResource(t *testing.T, res map[string]any) (*unstructured.Unstructured, *resource.Resource) {
```
jveski, line 89: controller_test.go had only these two helpers so I moved them into helpers_test.go

internal/controllers/reconciliation/helpers_test.go
```diff
@@ -52,3 +58,42 @@ func writeComposition(t *testing.T, client client.Client, orphan bool) (*apiv1.S
...
+func setupTestSubject(t *testing.T, mgr *testutil.Manager) {
+	rswb := flowcontrol.NewResourceSliceWriteBufferForManager(mgr.Manager)
+
+	var cache resource.Cache
+	rateLimiter := workqueue.DefaultTypedItemBasedRateLimiter[resource.Request]()
+	queue := workqueue.NewTypedRateLimitingQueue(rateLimiter)
+	cache.SetQueue(queue)
+
+	downstream := rest.CopyConfig(mgr.DownstreamRestConfig)
+	downstream.QPS = 200 // minimal throttling for the tests
```
jveski, line 71: This and the line above are the only changes from the current code

chihshenghuang (approved):

PR #290, opened 2025-02-21
More intuitive latency reporting for write buffer

jveski:
Currently the write buffer reports the latency as the period between the latest item being added to the buffer and it being flushed to the server.

This doesn't map to what I (most people?) would expect, which is the time between the _first_ item being added to the buffer and it being flushed.

So the fix is to just make the insertion time follow the queue item lifecycle: set it when initially enqueuing, remove it when forgetting the rate limit.

### DIFF
--- a/internal/flowcontrol/writebuffer.go
+++ b/internal/flowcontrol/writebuffer.go
@@ -72,7 +72,9 @@ func (w *ResourceSliceWriteBuffer) PatchStatusAsync(ctx context.Context, ref *re
 		}
 	}
 
-	w.insertionTime[key] = time.Now()
+	if _, found := w.insertionTime[key]; !found {
+		w.insertionTime[key] = time.Now()
+	}
 	w.state[key] = append(currentSlice, &resourceSliceStatusUpdate{
 		SlicedResource: ref,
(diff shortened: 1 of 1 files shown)


internal/flowcontrol/writebuffer.go
```diff
@@ -113,15 +115,14 @@ func (w *ResourceSliceWriteBuffer) processQueueItem(ctx context.Context) bool {
...
-	} else {
-		delete(w.insertionTime, sliceNSN)
 	}
 	w.mut.Unlock()
 
 	// We only forget the rate limit once the update queue for this slice is empty.
 	// So the first write is fast, but a steady stream of writes will be throttled exponentially.
 	if len(updates) == 0 {
 		w.queue.Forget(item)
+		delete(w.insertionTime, sliceNSN)
```
chihshenghuang, line 125: should this delete run in mutex lock also? the len(updates) might be different outside the mutex block

jveski: good catch! Fixed and confirmed that the tests pass with the race detector enabled

chihshenghuang (approved):

PR #291, opened 2025-02-21
Merge reconstitution controller into the reconciliation controller

jveski:
Move the reconstitution code into a controller runtime Source implementation used by the reconciliation controller - it's easier to track for folks who are already familiar with typical controller patterns and will make more sense as we expand test coverage of the reconciliation controller.


internal/controllers/reconciliation/controller.go
```diff
@@ -124,13 +117,7 @@ func (c *Controller) Reconcile(ctx context.Context, req resource.Request) (ctrl.
 	// Find the current and (optionally) previous desired states in the cache
 	var prev *resource.Resource
 	resource, visible, exists := c.resourceClient.Get(ctx, comp.Status.GetCurrentSynthesisUUID(), req.Resource)
-	if !exists {
-		// Returning is safe because filling the cache will enqueue a new work item
-		logger.V(1).Info("dropping work item because the corresponding synthesis no longer exists in the cache")
-		return ctrl.Result{}, nil
-	}
-	if !visible {
-		logger.V(1).Info("skipping because the resource isn't visible yet")
```
jveski, line 133: These logs are too chirpy and I've personally never found them useful. Open to keeping them but I tend to favor less log noise

tilnl (approved):

internal/controllers/reconciliation/controller.go
```diff
@@ -170,8 +157,6 @@ func (c *Controller) Reconcile(ctx context.Context, req resource.Request) (ctrl.
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	// If we modified the resource, we should also re-evaluate readiness
```
chihshenghuang, line 173: why remove this comment?

PR #293, opened 2025-02-24
Improved resource interface

jveski:
- Don't bother caching the JSON manifest since we have the parsed SMD representation anyway
- Don't read the resource slice from the reconciliation controller - move the cached status up from the private indexed resource struct to the public Resource type

### DIFF
--- a/internal/resource/resource.go
+++ b/internal/resource/resource.go
@@ -241,10 +241,6 @@ func NewResource(ctx context.Context, slice *apiv1.ResourceSlice, index int) (*R
 	if err != nil {
 		return nil, fmt.Errorf("invalid json: %w", err)
 	}
-	if parsed.Object != nil {
-		delete(parsed.Object, "status")
-		parsed.SetCreationTimestamp(metav1.Time{})
-	}
 
 	res.value = value.NewValueInterface(parsed.Object)
 	gvk := parsed.GroupVersionKind()
--- a/internal/resource/resource.go
+++ b/internal/resource/resource.go
@@ -241,6 +241,10 @@ func NewResource(ctx context.Context, slice *apiv1.ResourceSlice, index int) (*R
 	if err != nil {
 		return nil, fmt.Errorf("invalid json: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


internal/resource/resource.go
```diff
@@ -249,7 +247,7 @@ func NewResource(ctx context.Context, slice *apiv1.ResourceSlice, index int) (*R
 	logger := logr.FromContextOrDiscard(ctx)
 	resource := slice.Spec.Resources[index]
```
AYM1607, line 248: Unrelated to this PR, but maybe different naming would be beneficial here to differentiate the resource from the slice from our internal representation.

internal/resource/resource.go
```diff
@@ -259,10 +257,20 @@ func NewResource(ctx context.Context, slice *apiv1.ResourceSlice, index int) (*R
...
+	hash.Write([]byte(resource.Manifest))
+	res.ManifestHash = hash.Sum(nil)
+
+	parsed := &unstructured.Unstructured{}
+	err := parsed.UnmarshalJSON([]byte(resource.Manifest))
 	if err != nil {
 		return nil, fmt.Errorf("invalid json: %w", err)
 	}
+	if parsed.Object != nil {
+		delete(parsed.Object, "status")
```
AYM1607, line 270: Reasoning behind this? A comment would be great

tilnl (approved):

internal/resource/resource.go
jveski: I'm thinking we can just remove this code now that we have the resource version check? It protected against a case where the synthesizer produces resources that have non-nil status structs since they would never converge if the resource kind uses the status subresource.

jveski: Well `go-synth-example` is failing so apparently I'm missing something lol

jveski: I think the test was only failing because I didn't rebase on the commit that includes the resource version check. But regardless let's leave this as-is in this PR and maybe removed it later. Added a comment.

AYM1607 (approved):

PR #294, opened 2025-02-24
Handle case where synthesizer sees newer inputs than watch controller

jveski:
It's possible for the watch controller or its informers to lag behind the view of the input resources seen by synthesizer pods. In this case the version equality check in the scheduling controller will be false and cause a resynthesis loop until the watch controller catches up.

### DIFF
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -1,6 +1,8 @@
 package v1
 
 import (
+	"strconv"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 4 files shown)


api/v1/composition.go
```diff
@@ -121,17 +123,25 @@ type InputRevisions struct {
 	SynthesizerGeneration *int64 `json:"synthesizerGeneration,omitempty"`
 }
 
-func (i *InputRevisions) Equal(b InputRevisions) bool {
+func (i *InputRevisions) Less(b InputRevisions) bool {
 	if i.Key != b.Key {
-		return false
+		return false // shouldn't be possible
```
AYM1607, line 128: Panic?

chihshenghuang (approved):

api/v1/composition.go
```diff
@@ -121,17 +123,25 @@ type InputRevisions struct {
...
 	}
 	if (i.Revision == nil) != (b.Revision == nil) {
-		return false
+		return true
```
AYM1607, line 131: Should this always be true? It feels off for this to pass when b == nil, but i is not

jveski: My thinking was to fail open and just swap in the case that we can't determine ordering (one has revision one doesn't, etc.). But thinking more about it it's probably safer to fall back to the resource version. Thoughts?

AYM1607: I'm okay with failing open here as long as we document the behavior. Though we don't have any logging here so potential issues could go unseen. Either way works for me.

AYM1607 (approved):

PR #296, opened 2025-03-04
Unblock reconciliation during synthesis

### LINKED ISSUE #229: Store a third synthesis
Each composition has a max of two syntheses - current and previous. This allows Eno to compute three-way patches but only when the current synthesis has succeeded. So the reconciliation loop is broken during synthesis.

This is fine for most cases, but ideally reconciliation could continue regardless of synthesis state to handle cases where synthesis fails terminally.

jveski:
Resolves #229.

Since Eno only currently retains two syntheses, it isn't possible to (correctly) reconcile resources while synthesis is in-progress. So configuration might drift in cases where synthesis is slow.

It isn't a huge problem currently since synthesis is fast, but keeping a third synthesis ends up significantly simplifying the synthesis controller, which I'm working on property-based tests and refactoring for currently.

This change adds an "in flight" synthesis that is swapped into the existing "current" slot when synthesis is complete. I tried not to touch unrelated code - the diff mostly just swaps synthesis references, with a few exceptions (I'll leave PR comments to clarify).

## Backcompat

- Compositions with `CurrentSynthesis.Synthesized != nil` are unaffected - this is still the terminal state
- Compositions with `CurrentSynthesis == nil` are unaffected since the new code will be used for synthesis
- Compositions with `CurrentSynthesis != nil && CurrentSynthesis.Synthesized == nil` (a.k.a. in-flight synthesis) will converge eventually because the scheduling controller considers `(CurrentSynthesis.Synthesized == nil) == (CurrentSynthesis == nil)`

### DIFF
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -52,7 +52,7 @@ type CompositionSpec struct {
 
 type CompositionStatus struct {
 	Simplified         *SimplifiedStatus `json:"simplified,omitempty"`
-	PendingSynthesis   *Synthesis        `json:"pendingSynthesis,omitempty"`
+	InFlightSynthesis  *Synthesis        `json:"inFlightSynthesis,omitempty"`
 	CurrentSynthesis   *Synthesis        `json:"currentSynthesis,omitempty"`
 	PreviousSynthesis  *Synthesis        `json:"previousSynthesis,omitempty"`
 	InputRevisions     []InputRevisions  `json:"inputRevisions,omitempty"`
@@ -225,7 +225,7 @@ func (c *Composition) ShouldIgnoreSideEffects() bool {
 }
 
 func (c *Composition) Synthesizing() bool {
-	return c.Status.PendingSynthesis != nil
+	return c.Status.InFlightSynthesis != nil
 }
 
 func (c *Composition) EnableIgnoreSideEffects() {
@@ -258,8 +258,8 @@ func (c *Composition) ShouldOrphanResources() bool {
 }
 
 func (s *CompositionStatus) getLatestSynthesisUUID() string {
-	if s.PendingSynthesis != nil {
-		return s.PendingSynthesis.UUID
+	if s.InFlightSynthesis != nil {
+		return s.InFlightSynthesis.UUID
 	}
 	if s.CurrentSynthesis != nil {
 		return s.CurrentSynthesis.UUID
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -214,14 +214,6 @@ func (c *Composition) InputsOutOfLockstep(synth *Synthesizer) bool {
 }
 
 func (s *CompositionStatus) GetCurrentSynthesisUUID() string {
-	if s.CurrentSynthesis == nil {
-		return ""
-	}
-	return s.CurrentSynthesis.UUID
-}
-
-// TODO: Remove other?
-func (s *CompositionStatus) GetLatestSynthesisUUID() string {
 	if s.PendingSynthesis != nil {
 		return s.PendingSynthesis.UUID
 	}
@@ -255,13 +247,13 @@ func (c *Composition) ForceResynthesis() {
 	if anno == nil {
 		anno = map[string]string{}
 	}
-	anno[forceResynthesisAnnotation] = c.Status.GetLatestSynthesisUUID()
+	anno[forceResynthesisAnnotation] = c.Status.GetCurrentSynthesisUUID()
 	c.SetAnnotations(anno)
 }
 
 func (c *Composition) ShouldForceResynthesis() bool {
 	val, ok := c.GetAnnotations()[forceResynthesisAnnotation]
-	return ok && val == c.Status.GetLatestSynthesisUUID()
+	return ok && val == c.Status.GetCurrentSynthesisUUID()
 }
 
 func (c *Composition) ShouldOrphanResources() bool {
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -214,6 +214,14 @@ func (c *Composition) InputsOutOfLockstep(synth *Synthesizer) bool {
 }
 
 func (s *CompositionStatus) GetCurrentSynthesisUUID() string {
+	if s.CurrentSynthesis == nil {
+		return ""
+	}
+	return s.CurrentSynthesis.UUID
+}
+
+// TODO: Remove other?
+func (s *CompositionStatus) GetLatestSynthesisUUID() string {
 	if s.PendingSynthesis != nil {
 		return s.PendingSynthesis.UUID
 	}
@@ -247,13 +255,13 @@ func (c *Composition) ForceResynthesis() {
 	if anno == nil {
 		anno = map[string]string{}
 	}
-	anno[forceResynthesisAnnotation] = c.Status.GetCurrentSynthesisUUID()
+	anno[forceResynthesisAnnotation] = c.Status.GetLatestSynthesisUUID()
 	c.SetAnnotations(anno)
 }
 
 func (c *Composition) ShouldForceResynthesis() bool {
 	val, ok := c.GetAnnotations()[forceResynthesisAnnotation]
-	return ok && val == c.Status.GetCurrentSynthesisUUID()
+	return ok && val == c.Status.GetLatestSynthesisUUID()
 }
 
 func (c *Composition) ShouldOrphanResources() bool {
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -220,17 +220,6 @@ func (s *CompositionStatus) GetCurrentSynthesisUUID() string {
 	return s.CurrentSynthesis.UUID
 }
 
-// TODO: Remove other?
-func (s *CompositionStatus) GetLatestSynthesisUUID() string {
-	if s.PendingSynthesis != nil {
-		return s.PendingSynthesis.UUID
-	}
-	if s.CurrentSynthesis != nil {
-		return s.CurrentSynthesis.UUID
-	}
-	return ""
-}
-
 func (c *Composition) ShouldIgnoreSideEffects() bool {
 	return c.Annotations["eno.azure.io/ignore-side-effects"] == "true"
 }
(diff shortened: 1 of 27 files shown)


api/v1/composition.go
```diff
@@ -52,6 +52,7 @@ type CompositionSpec struct {
 
 type CompositionStatus struct {
 	Simplified         *SimplifiedStatus `json:"simplified,omitempty"`
+	PendingSynthesis   *Synthesis        `json:"pendingSynthesis,omitempty"`
```
jveski, line 55: Does "pending" make sense here? Maybe "Active"? or "InFlight"?

internal/controllers/reconciliation/crud_test.go
```diff
@@ -702,59 +702,6 @@ func TestDisableUpdates(t *testing.T) {
 	assert.Equal(t, "baz", obj.Data["foo"])
 }
 
-// TestOrphanedCompositionDeletion proves that compositions can be deleted when their synthesizer is missing.
-func TestOrphanedCompositionDeletion(t *testing.T) {
```
jveski, line 706: The logic that made this test pass would be... tricky to get working after this change. That said, I don't personally really care about the behavior at least until some other refactoring is done. Open to commenting out the test with a TODO if folks would prefer, but ultimately I don't think it's the right place to test this behavior anyway.

internal/controllers/scheduling/controller.go
```diff
@@ -161,9 +161,10 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 func (c *controller) getNextCooldownSlot(comps *apiv1.CompositionList) time.Time {
 	var next time.Time
 	for _, comp := range comps.Items {
-		syn := comp.Status.CurrentSynthesis
-		if syn != nil && syn.Deferred && syn.Initialized != nil && syn.Initialized.Time.After(next) {
-			next = syn.Initialized.Time
+		for _, syn := range []*apiv1.Synthesis{comp.Status.PendingSynthesis, comp.Status.CurrentSynthesis, comp.Status.PreviousSynthesis} {
```
jveski, line 164: It was kind of a bug to not check PreviousSynthesis before 🤷‍♂️

internal/controllers/synthesis/lifecycle.go
```diff
@@ -198,37 +199,16 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 func (c *podLifecycleController) reconcileDeletedComposition(ctx context.Context, comp *apiv1.Composition) (ctrl.Result, error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
-	// If the composition was being synthesized at the time of deletion we need to swap the previous
-	// state back to current. Otherwise we'll get stuck waiting for a synthesis that can't happen.
-	if shouldRevertStateSwap(comp) {
```
jveski, line 203: We don't need this any more!

api/v1/composition.go
chihshenghuang: I like "InFlight" :grin:

jveski: Done!

internal/controllers/reconciliation/crud_test.go
AYM1607: I'm okay with deleting this and opening up an issue that points to this PR if we care enough about the behavior to test it after the refactor.

api/v1/composition.go
```diff
@@ -224,7 +225,7 @@ func (c *Composition) ShouldIgnoreSideEffects() bool {
 }
 
 func (c *Composition) Synthesizing() bool {
-	return c.Status.CurrentSynthesis != nil && c.Status.CurrentSynthesis.Synthesized == nil
+	return c.Status.InFlightSynthesis != nil
```
AYM1607, line 228: Clean 😼

docs/api.md
```diff
@@ -92,6 +92,7 @@ _Appears in:_
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
 | `simplified` _[SimplifiedStatus](#simplifiedstatus)_ |  |  |  |
+| `inFlightSynthesis` _[Synthesis](#synthesis)_ |  |  |  |
```
AYM1607, line 95: Maybe we should start adding more comments to the api structs so we can use this auto-gen files as proper docs

internal/controllers/synthesis/lifecycle.go
```diff
@@ -68,8 +69,8 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
 
 	logger = logger.WithValues("compositionName", comp.Name,
 		"compositionNamespace", comp.Namespace,
-		"compositionGeneration", comp.Generation,
-		"synthesisID", comp.Status.GetCurrentSynthesisUUID())
```
AYM1607, line 72: Replace with get latest uuid?

api/v1/config/crd/eno.azure.io_synthesizers.yaml
```diff
@@ -350,7 +350,7 @@ spec:
                                         pod labels will be ignored. The default value is empty.
                                         The same key is forbidden to exist in both matchLabelKeys and labelSelector.
                                         Also, matchLabelKeys cannot be set when labelSelector isn't set.
-                                        This is an alpha field and requires enabling MatchLabelKeysInPodAffinity feature gate.
+                                        This is a beta field and requires enabling MatchLabelKeysInPodAffinity feature gate (enabled by default).
```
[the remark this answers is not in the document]
jveski: I must have updated controller-gen on my devbox recently 😆

internal/controllers/synthesis/lifecycle.go
AYM1607: I see, it's not relevant until we know that we're acting upon the in flight uuid.

docs/api.md
jveski: Agreed - we're pretty close on a few structs but most of them are still lacking

internal/controllers/synthesis/lifecycle.go
```diff
@@ -136,15 +136,17 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
 	}
+	logger = logger.WithValues("synthesisID", comp.Status.InFlightSynthesis.UUID)
+	ctx = logr.NewContext(ctx, logger)
 
 	// Back off to avoid constantly re-synthesizing impossible compositions (unlikely but possible)
 	if shouldBackOffPodCreation(comp) {
 		const base = time.Millisecond * 250
-		wait := base * time.Duration(comp.Status.CurrentSynthesis.Attempts)
-		nextAttempt := comp.Status.CurrentSynthesis.PodCreation.Time.Add(wait)
+		wait := base * time.Duration(comp.Status.InFlightSynthesis.Attempts)
```
AYM1607, line 148: I think we had this conversation already, but could you remind me why we're not doing exponential here?

internal/controllers/scheduling/controller.go
```diff
@@ -161,9 +161,10 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 func (c *controller) getNextCooldownSlot(comps *apiv1.CompositionList) time.Time {
 	var next time.Time
 	for _, comp := range comps.Items {
-		syn := comp.Status.CurrentSynthesis
-		if syn != nil && syn.Deferred && syn.Initialized != nil && syn.Initialized.Time.After(next) {
-			next = syn.Initialized.Time
+		for _, syn := range []*apiv1.Synthesis{comp.Status.InFlightSynthesis, comp.Status.CurrentSynthesis, comp.Status.PreviousSynthesis} {
```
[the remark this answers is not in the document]
jveski: IMO it's _kind of_ a bug to not check prev. But realistically synthesis is slow enough relative to the cooldown period that it wouldn't ever matter.

internal/controllers/synthesis/lifecycle.go
jveski: No reason that I'm aware of lol

I have a half working PR that integrates the scheduling controller into the retry loop so retries will be prioritized correctly. That one will also switch to exponential backoff.

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -147,16 +147,22 @@ func (c *cleanupDecision) String() string {
 }
 
 func shouldDeleteSlice(comp *apiv1.Composition, slice *apiv1.ResourceSlice) bool {
-	if comp.Status.CurrentSynthesis == nil || slice.Spec.CompositionGeneration > comp.Status.CurrentSynthesis.ObservedCompositionGeneration {
+	var synthesizedGeneration int64
+	if comp.Status.InFlightSynthesis != nil {
```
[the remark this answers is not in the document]
jveski: `InFlightSynthesis.Synthesized == true` is not currently possible since in-flight is always swapped to current synthesis by the executor. We can probably remove the property actually, just need to clean up the logs that depend on it

internal/controllers/synthesis/slicecleanup.go
```diff
@@ -147,16 +147,22 @@ func (c *cleanupDecision) String() string {
...
-		isReferencedByComp = synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) || synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice)
-		isSynthesized      = comp.Status.CurrentSynthesis.Synthesized != nil
+		hasBeenRetried     = slice.Spec.Attempt != 0 && comp.Status.InFlightSynthesis != nil && comp.Status.InFlightSynthesis.Attempts > slice.Spec.Attempt && slice.Spec.SynthesisUUID == comp.Status.InFlightSynthesis.UUID
+		isReferencedByComp = synthesisReferencesSlice(comp.Status.InFlightSynthesis, slice) || synthesisReferencesSlice(comp.Status.CurrentSynthesis, slice) || synthesisReferencesSlice(comp.Status.PreviousSynthesis, slice)
+		isSynthesized      = comp.Status.CurrentSynthesis != nil
 		compIsDeleted      = comp.DeletionTimestamp != nil
-		fromOldComposition = slice.Spec.CompositionGeneration < comp.Status.CurrentSynthesis.ObservedCompositionGeneration
+		fromOldComposition = comp.Status.CurrentSynthesis == nil || slice.Spec.CompositionGeneration < comp.Status.CurrentSynthesis.ObservedCompositionGeneration
```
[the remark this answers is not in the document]
jveski: It means that the resource slice was derived from a composition generation that is < the current

internal/controllers/selfhealing/slice_test.go
```diff
@@ -480,14 +480,14 @@ func TestNotEligibleForResynthesis(t *testing.T) {
 		{
 			name: "CurrentSynthesis Synthesized is nil",
```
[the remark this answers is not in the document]
AYM1607: Sure thing!

internal/controllers/synthesis/slicecleanup.go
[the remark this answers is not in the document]
jveski: Hmm this might deserve more scrutiny actually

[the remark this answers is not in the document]
jveski: Pushed some tests to help reason about this. I _think_ this effectively preserves the current logic, although I'm pretty sure the current logic doesn't make any sense

AYM1607 (approved):

PR #297, opened 2025-03-06
Add fuzz-style test framework

jveski:
Eno already has some property-based testing (style) tests in the scheduling package, but there's a lot of boilerplate involved and some indirection would be helpful if we're going to expand the number of tests significantly.

This framework essentially just re-implements the existing boilerplate but slightly nicer. I'll update the test case to use it in another PR.

### DIFF
--- a/internal/testutil/statespace/statespace.go
+++ b/internal/testutil/statespace/statespace.go
@@ -1,0 +1,98 @@
+package statespace
+
+import (
+	"math/rand/v2"
+	"strings"
+	"testing"
+)
+
+// Model implements a simple fuzz-style test framework.
+// Rather than fuzz scalar values like property-based tests, this framework enumerates the subsets of a bounded, high-cardinality state space.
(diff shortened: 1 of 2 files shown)


internal/testutil/statespace/statespace.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"math/rand/v2"
+	"strings"
+	"testing"
+)
+
+// Model implements a simple fuzz-style test framework.
+// Rather than fuzz scalar values like property-based tests, this framework enumerates the subsets of a bounded, high-cardinality state space.
+// Essentially, the framework will call the subject with every subset of the known mutations and assert on each invariant.
+// Application order of mutations is randomized to avoid order-dependence without fully permuting the space (the implementation is very fast).
```
jveski, line 12: Do these comments make any sense? 🤣 

We're not _quite_ doing property-based testing but I'm not aware of a better term.

PR #299, opened 2025-03-12
Extract pod garbage collection into a separate controller

jveski:
Currently the synthesis lifecycle controller creates and deletes pods. But since it reconciles compositions, it needs to reconcile the entire set of pods associated with the composition for every reconciliation. This works, but it's confusing to read/test, and it makes requeing to check timeouts tricky.

This PR refactors the deletion logic into a separate controller, extends test coverage around some important edge cases, and simplifies the existing lifecycle controller somewhat. Mostly it unblocks better testing of the lifecycle controller in the future.


internal/controllers/synthesis/integration_test.go
```diff
@@ -238,3 +242,38 @@ func TestControllerSwitchingSynthesizers(t *testing.T) {
 		assert.NotEqual(t, comp.Status.CurrentSynthesis.ResourceSlices, initialSlices)
 	})
 }
+
+func TestControllerBackoff(t *testing.T) {
```
jveski, line 246: Unrelated to the PR: moved from backoff_test.go since it was the only test in the file

internal/controllers/synthesis/lifecycle.go
```diff
@@ -197,6 +150,41 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
+		}
+
+		if podIsCurrent(comp, &pod) {
+			if !active {
+				active = true
+				continue
+			}
+
+			// Clean up any dupes
+			err := c.client.Delete(ctx, &pod)
```
jveski, line 176: There is one deletion case that can't logically be addressed by the podGC controller

internal/controllers/synthesis/lifecycle.go
```diff
@@ -197,6 +166,41 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
+
+	var active bool
+	var deleting bool
+	for _, pod := range pods.Items {
+		if pod.DeletionTimestamp != nil {
+			if !deleting {
+				deleting = true
+				continue
+			}
+			return false, nil // don't create new pod if two are deleting
```
AYM1607, line 182: Mind expanding on why we adopt this heuristic?

internal/controllers/synthesis/lifecycle.go
AYM1607: It feels weird to have this here, why can't it be done in the gc controller?

internal/controllers/synthesis/gc.go
```diff
@@ -0,0 +1,156 @@
...
+		logger = logger.WithValues("reason", "SynthesizerDeleted")
+		return ctrl.Result{}, p.deletePod(ctx, pod, logger)
+	}
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("getting synthesizer: %w", err)
+	}
+
+	// Pods from old syntheses
+	if !podIsCurrent(comp, pod) {
+		const gracePeriod = 250 * time.Millisecond
```
AYM1607, line 88: Could informer drift be larger than this in underlays?

internal/controllers/synthesis/lifecycle.go
jveski: Hmm let me see what the tradeoffs would be

internal/controllers/synthesis/lifecycle.go
```diff
@@ -83,64 +79,37 @@ func (c *podLifecycleController) Reconcile(ctx context.Context, req ctrl.Request
...
+	})
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("listing pods: %w", err)
 	}
 
-	// Synthesis isn't possible without a synth
-	if syn == nil {
-		return ctrl.Result{}, nil
+	ctx = logr.NewContext(ctx, logger)
+	ok, err := c.safeToCreatePod(ctx, pods, comp)
```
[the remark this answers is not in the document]
jveski: Exactly - we check the backoff interval first to avoid making the apiserver call before the next retry is ready

internal/controllers/synthesis/gc.go
jveski: Let me see if there's a better way to handle this, or at least add some comments

internal/controllers/synthesis/lifecycle.go
jveski: Will try to refactor a bit for clarity here

internal/controllers/synthesis/gc.go
jveski: Up'd it to 1 second just in case

jveski: My take is that it's extremely unlikely informers will be behind by >1s and even if they are, recreating the pod isn't a terrible failure mode.

internal/controllers/synthesis/lifecycle.go
jveski: Remove it! Turns out we don't really need it anyway. I reworked a few things to avoid even needing a pod index.

AYM1607 (approved):
LGTM

PR #300, opened 2025-03-12
update fetchPreviousSlices func

chihshenghuang:
Fix the issue that previous slices is stored in current synthesis instead of previous synthesis before updating composition

### DIFF
--- a/internal/execution/executor_test.go
+++ b/internal/execution/executor_test.go
@@ -2,6 +2,8 @@ package execution
 
 import (
 	"context"
+	"slices"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/execution/executor_test.go
```diff
@@ -53,11 +55,11 @@ func TestBasics(t *testing.T) {
 				Object: map[string]any{
 					"apiVersion": "v1",
 					"kind":       "ConfigMap",
-					"metadata": map[string]string{
+					"metadata": map[string]any{
```
chihshenghuang, line 58: Need to pass metadata as type map[string]any or NestedString will throw the error

jveski (approved):

PR #302, opened 2025-03-14
Simplify resource slice controller

jveski:
There is a ton of history around the resource slice cleanup logic and it's gotten quite complicated.

- Create a new catch-all resource slice controller
- Merge in the existing slice cleanup controller
- Merge in the self healing controller
- Merge in the slice aggregation controller
- Significantly simplify the logic of the consolidated controllers


api/v1/resourceslice.go
```diff
@@ -20,9 +20,9 @@ type ResourceSlice struct {
 }
 
 type ResourceSliceSpec struct {
-	CompositionGeneration int64      `json:"compositionGeneration,omitempty"`
+	CompositionGeneration int64      `json:"compositionGeneration,omitempty"` // deprecated
```
AYM1607, line 23: 😮 apiserver does not reject requests with unregistered fields right?

AYM1607: For when we eventually remove these that is.

internal/controllers/resourceslice/slice.go
```diff
@@ -0,0 +1,197 @@
...
+		Complete(&sliceController{client: mgr.GetClient()})
+}
+
+func (s *sliceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
+	comp := &apiv1.Composition{}
+	err := s.client.Get(ctx, req.NamespacedName, comp)
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("getting composition: %w", err))
```
AYM1607, line 38: nit: Comment stating that deletion is handled by a different controller and some reasoning around why?

api/v1/resourceslice.go
jveski: Right- it returns warnings but not 400s

internal/controllers/resourceslice/slice.go
```diff
@@ -0,0 +1,197 @@
...
+		err := s.client.Get(ctx, client.ObjectKeyFromObject(slice), slice)
+		if errors.IsNotFound(err) {
+			if comp.DeletionTimestamp != nil {
+				logger.V(1).Info("resource slice is missing, ignoring because composition is being deleted", "resourceSliceName", ref.Name)
+				continue
+			}
+			return s.handleMissingSlice(ctx, comp, ref.Name)
+		}
+		if err != nil {
+			return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("getting resource slice: %w", err))
```
AYM1607, line 62: Is the IgnoreNotFound just defensive coding in case the logic above changes?

jveski: Nope, just copy paste 🙃

internal/controllers/resourceslice/slice.go
```diff
@@ -0,0 +1,197 @@
...
+		// Collect the state of every resource
+		for _, state := range slice.Status.Resources {
+			state := state
+			if resourceNotReconciled(comp, &state) {
+				snapshot.Reconciled = false
+			}
+			if state.Ready == nil {
+				snapshot.Ready = false
+			}
+			if state.Ready != nil && (snapshot.ReadyTime == nil || snapshot.ReadyTime.Before(state.Ready)) {
```
AYM1607, line 82: Is the aim here for snapshot to have the latest ready time? I think that it is, if so, do you think that `state.Ready.After(snapshot.ReadyTime)` would read better?

internal/controllers/resourceslice/slice.go
```diff
@@ -0,0 +1,197 @@
...
+	logger.V(1).Info("aggregated resource status into composition", "reconciled", snapshot.Reconciled, "ready", snapshot.Ready)
+
+	return ctrl.Result{}, nil
+}
+
+func (s *sliceController) handleMissingSlice(ctx context.Context, comp *apiv1.Composition, sliceName string) (ctrl.Result, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
+	// We can't do anything about missing resource slices if synthesis is already in-flight or it isn't safe to resynthesize
+	if comp.ShouldIgnoreSideEffects() || comp.Status.InFlightSynthesis != nil || comp.ShouldForceResynthesis() {
```
AYM1607, line 106: Why not `!comp.ShouldFroceResynthesis()` I guess I haven't really grasped the semantics around it.

internal/controllers/resourceslice/slice.go
```diff
@@ -0,0 +1,197 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	// It's possible that newly created slices haven't hit the informer cache yet
+	if synthd := comp.Status.CurrentSynthesis.Synthesized; synthd != nil {
+		delta := time.Since(synthd.Time)
+		if delta < time.Second*5 {
```
AYM1607, line 113: Kinda scary that we rely on wall clock time for this, but I guess there's no better way. _Should_ be okay because the synth time is written by the Eno controller and should be (in most scenarios barring leader election) the same wall clock.

AYM1607: Ah, it's also fine because there's a passthrough check below.

internal/controllers/resourceslice/slice.go
jveski: I like it 👍

internal/controllers/resourceslice/slice.go
jveski: That func is maybe not named so clearly. It's true when synthesis should be forced given the state of the composition. So in this case it's the idempotence check

internal/controllers/resourceslice/slice.go
jveski: This is just an optimization to avoid making the apiserver roundtrip that happens next :)

internal/controllers/resourceslice/slice.go
```diff
@@ -0,0 +1,197 @@
...
+			logger.V(0).Info("composition was reconciled", "latency", latency.Abs().Milliseconds())
+		}
+	} else {
+		comp.Status.CurrentSynthesis.Reconciled = nil
+	}
+
+	return true
+}
+
+// resourceNotReconciled returns true when a resource should be considered reconciled.
```
AYM1607, line 185: _not_ reconciled?

internal/controllers/resourceslice/slice.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	// Readiness
+	now := metav1.Now()
+	if snapshot.Ready && snapshot.ReadyTime != nil {
+		comp.Status.CurrentSynthesis.Ready = snapshot.ReadyTime
+
+		if synthed := comp.Status.CurrentSynthesis.Synthesized; synthed != nil {
+			latency := snapshot.ReadyTime.Sub(synthed.Time)
+			if latency.Milliseconds() > 0 {
```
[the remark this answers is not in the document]
jveski: i.e. comps with `CurrentSynthesis.Synthesized == nil`? That shouldn't actually ever be possible so yeah probably makes sense to log something obvious if it were ever to happen

internal/controllers/resourceslice/slicecleanup.go
```diff
@@ -0,0 +1,192 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&apiv1.ResourceSlice{}).
+		WatchesRawSource(source.Kind(mgr.GetCache(), &apiv1.Composition{}, c.newCompEventHandler())).
+		WithLogConstructor(manager.NewLogConstructor(mgr, "sliceCleanupController")).
+		Complete(c)
+}
+
+func (c *cleanupController) newCompEventHandler() handler.TypedEventHandler[*apiv1.Composition, reconcile.Request] {
+	fn := func(c *apiv1.Composition, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+		for _, syn := range []*apiv1.Synthesis{c.Status.CurrentSynthesis, c.Status.PreviousSynthesis} {
```
AYM1607, line 49: Is there a scenario where we could miss something from inflight?

internal/controllers/resourceslice/slice.go
jveski: nice catch!

internal/controllers/resourceslice/slicecleanup.go
jveski: Good question. I don't think so - the inflight synthesis never actually has any resource slices. They're added right before it's swapped into the current slot.

internal/controllers/resourceslice/slice.go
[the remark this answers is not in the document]
AYM1607: I meant the latency log for when len(slices) == 0, since we'd never see it in that chase but would still become synthesized/ready

internal/controllers/resourceslice/slicecleanup.go
AYM1607: But those slices _could_ still exist even if in flight never gets swapped (could that ever happen?), do we clean them up in that case?

internal/controllers/resourceslice/slice.go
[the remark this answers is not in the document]
jveski: Ahh gotcha - that's a good point. Fixed, also did some refactoring to make this bit more legible.

internal/controllers/resourceslice/slicecleanup.go
jveski: The same write that adds the resource slice refs also swaps the states, so it shouldn't be possible. That said- I could imagine this changing in the future so let's just add in-flight to the slice just in case.

AYM1607: Could it be the case that slices are created but never refd? In any case, I think checking InFlight, should take care of it as no slice is ever not associated to a synthesis.

AYM1607 (approved):
LGTM

internal/controllers/resourceslice/slicecleanup.go
jveski: It's possible they wouldn't be ref'd but in that case checking the in-flight synthesis won't actually enqueue anything. It's okay though since resource slice creation enqueues anyway.

PR #303, opened 2025-03-17
Re-enter the scheduling controller on synthesis timeout

jveski:
Currently synthesis pods are recreated using exponential backoff until they succeed or fail. But the retries are handled by the synthesis lifecycle controller, which isn't aware of the priorities/limits imposed by the scheduling controller.

So impossible syntheses will count against the concurrency limit forever - potentially blocking other more important operations.

This change replaces the pod-level retries with synthesis-scoped retries, which essentially re-enter the scheduling controller at a lower priority.

### DIFF
--- a/internal/controllers/aggregation/composition.go
+++ b/internal/controllers/aggregation/composition.go
@@ -72,6 +72,8 @@ func (c *compositionController) aggregate(synth *apiv1.Synthesizer, comp *apiv1.
 		return copy
 	}
 
+	// TODO: New status for synthesis retry backoff
+
 	copy.Status = "Synthesizing"
 	if !comp.InputsExist(synth) {
 		copy.Status = "MissingInputs"
--- a/internal/controllers/composition/controller.go
+++ b/internal/controllers/composition/controller.go
@@ -1,0 +1,123 @@
+package composition
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/google/uuid"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	apiv1 "github.com/Azure/eno/api/v1"
+	"github.com/Azure/eno/internal/manager"
+)
+
+type compositionController struct {
+	client client.Client
+}
+
+func NewController(mgr ctrl.Manager) error {
+	c := &compositionController{
+		client: mgr.GetClient(),
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&apiv1.Composition{}).
+		WithLogConstructor(manager.NewLogConstructor(mgr, "compositionController")).
+		Complete(c)
+}
+
+func (c *compositionController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+	comp := &apiv1.Composition{}
+	err := c.client.Get(ctx, req.NamespacedName, comp)
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(fmt.Errorf("getting composition resource: %w", err))
+	}
+
+	logger = logger.WithValues("compositionName", comp.Name, "compositionNamespace", comp.Namespace, "compositionGeneration", comp.Generation, "synthesisID", comp.Status.GetCurrentSynthesisUUID())
+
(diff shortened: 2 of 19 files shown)


internal/controllers/aggregation/composition.go
```diff
@@ -72,6 +72,8 @@ func (c *compositionController) aggregate(synth *apiv1.Synthesizer, comp *apiv1.
 		return copy
 	}
 
+	// TODO: New status for synthesis retry backoff
```
AYM1607, line 75: Will this be addressed in this PR?

internal/controllers/scheduling/controller.go
```diff
@@ -129,8 +129,10 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if op == nil || inFlight >= c.concurrencyLimit {
 		return ctrl.Result{}, nil
 	}
-	if wait := time.Until(nextSlot); op.Reason.Deferred() && wait > 0 {
-		return ctrl.Result{RequeueAfter: wait}, nil
+	if !op.NotBefore.IsZero() {
```
AYM1607, line 132: This is a bit criptic, what is it supposed to match on?

internal/controllers/composition/controller.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	// Enforce the synthesis timeout period
+	if syn := comp.Status.InFlightSynthesis; syn != nil && syn.Initialized != nil && synth.Spec.PodTimeout != nil {
+		delta := time.Until(syn.Initialized.Time.Add(synth.Spec.PodTimeout.Duration))
+		if delta > 0 {
+			return ctrl.Result{RequeueAfter: delta}, nil
+		}
+		syn.Canceled = ptr.To(metav1.Now())
+		if err := c.client.Status().Update(ctx, comp); err != nil {
```
AYM1607, line 74: Worth patching here to avoid conflict? I'm not sure if you're relying conflict errors for anything here though

internal/controllers/composition/controller.go
```diff
@@ -0,0 +1,123 @@
...
+			comp.Status.CurrentSynthesis.Ready = nil
+			err := c.client.Status().Update(ctx, comp)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("updating current composition generation: %w", err)
+			}
+			logger.V(0).Info("updated composition status to reflect deletion", "synthesisID", comp.Status.CurrentSynthesis.UUID)
+			return ctrl.Result{}, nil
+		}
+
+		if syn.Reconciled == nil {
```
AYM1607, line 107: The reconciler sets this to non-nil (because it has no slices) after the status update above right?

internal/controllers/scheduling/op.go
```diff
@@ -19,20 +19,34 @@ type op struct {
...
 	o.Reason, ok = classifyOp(synth, comp)
 	if !ok {
-		return nil
+		syn := comp.Status.InFlightSynthesis
+		if syn == nil || syn.Canceled == nil {
+			return nil
+		}
+
+		// Retry canceled syntheses with exponential backoff
+		backoff := time.Duration(float64(syn.Canceled.Sub(syn.Initialized.Time)) * float64(syn.Attempts) * 1.5)
```
AYM1607, line 42: Why not use the pod timeout instead of cancelled - initialized?

internal/controllers/aggregation/composition.go
jveski: I am planning on folding the status aggregation logic into the composition controller but didn't want to make this PR any larger. So I'll follow up once this merges.

internal/controllers/scheduling/op.go
jveski: This is completely pedantic but I was thinking about what happens if the synth's pod timeout changes. I'd expect existing syntheses to back off using the timeout specified by their `observedSynthesizerGeneration` - not the new value.

internal/controllers/composition/controller.go
jveski: Yep (the resource slice controller though)

internal/controllers/composition/controller.go
jveski: I don't think it's a common enough occurrence to matter. But mostly I'm just afraid of coming up with the wrong `test` ops and making more weird concurrency bugs

internal/controllers/scheduling/controller.go
jveski: lol yes this does require a minute to compute. Added a comment.

internal/controllers/scheduling/op.go
AYM1607: That's a valid reason, I was just curious about why one over the other. A (welcome?) side effect of this is that it might add a bit of jitter.

AYM1607 (approved):
LGTM

PR #310, opened 2025-03-19
feat: expose chart loader options to caller

bcho:
In our use case, we want to embed the helm chart into the binary via embedding.FS . In this case, we want to provide the loaded chart (based on helm's https://github.com/helm/helm/blob/fa9efb07d9d8debbb4306d72af76a383895aa8c4/pkg/chart/loader/load.go#L72 ) This pull request refactored the ChartPath option to ChartLoader for this support.


jveski (approved):
🚢

PR #311, opened 2025-03-19
Merge/simplify the symphony aggregation and replication controllers

jveski:
It doesn't really make sense for these controllers to be separate - there's a lot of shared logic between them.

This essentially just merges the two controllers with a bit of refactoring and better tests.


api/v1/symphony.go
```diff
@@ -46,7 +46,7 @@ type SymphonyStatus struct {
 	Synthesized        *metav1.Time     `json:"synthesized,omitempty"`
 	Reconciled         *metav1.Time     `json:"reconciled,omitempty"`
 	Ready              *metav1.Time     `json:"ready,omitempty"`
-	Synthesizers       []SynthesizerRef `json:"synthesizers,omitempty"`
+	Synthesizers       []SynthesizerRef `json:"synthesizers,omitempty"` // deprecated
```
jveski, line 49: This definitely didn't accomplish what it promised (lol)

I'll push another commit that replaces it soon

AYM1607 (approved):

PR #315, opened 2025-03-28
Support implicit bindings

jveski:
When a ref will only ever be bound to a single resource across all compositions, we can avoid the need to specify a binding by letting synthesizers declare the name/namespace.

### DIFF
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"fmt"
+	"slices"
 	"strconv"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -151,6 +152,17 @@ func (c *Composition) InputsExist(syn *Synthesizer) bool {
 	refs := map[string]struct{}{}
 	for _, ref := range syn.Spec.Refs {
 		refs[ref.Key] = struct{}{}
+
+		// Handle missing resources for implied bindings
+		if ref.Resource.Name == "" {
+			continue
+		}
+		found := slices.ContainsFunc(c.Status.InputRevisions, func(rev InputRevisions) bool {
+			return ref.Key == rev.Key
+		})
+		if !found {
+			return false
+		}
 	}
 
 	for _, binding := range c.Spec.Bindings {
@@ -159,14 +171,9 @@ func (c *Composition) InputsExist(syn *Synthesizer) bool {
 			// This is important for forwards compatibility- compositions can bind to refs that don't exist, but will in future synths
 			continue
 		}
-
-		var found bool
-		for _, rev := range c.Status.InputRevisions {
-			if binding.Key == rev.Key {
-				found = true
-				break
-			}
-		}
+		found := slices.ContainsFunc(c.Status.InputRevisions, func(rev InputRevisions) bool {
+			return binding.Key == rev.Key
+		})
 		if !found {
 			return false
 		}
--- a/api/v1/composition.go
+++ b/api/v1/composition.go
@@ -152,19 +152,9 @@ func (c *Composition) InputsExist(syn *Synthesizer) bool {
 	refs := map[string]struct{}{}
 	for _, ref := range syn.Spec.Refs {
 		refs[ref.Key] = struct{}{}
-
-		// Handle missing resources for implied bindings
-		if ref.Resource.Name == "" {
-			continue
-		}
-		found := slices.ContainsFunc(c.Status.InputRevisions, func(rev InputRevisions) bool {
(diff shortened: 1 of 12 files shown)


api/v1/composition.go
```diff
@@ -153,24 +154,40 @@ func (c *Composition) InputsExist(syn *Synthesizer) bool {
...
+				return ref.Key == rev.Key
+			})
+			if !found {
+				return false
 			}
+			continue
 		}
-		if !found {
+
+		// Every ref must be bound
```
jveski, line 185: This fixes an unrelated bug: refs without bindings should count as missing inputs

internal/controllers/composition/controller.go
```diff
@@ -143,15 +141,9 @@ func (c *compositionController) reconcileSimplifiedStatus(ctx context.Context, s
...
-	js, err := json.Marshal(&[]map[string]any{
-		// this is the only writer - no need for concurrency safety
-		{"op": "replace", "path": "/status/simplified", "value": next},
-	})
-	if err != nil {
-		return false, fmt.Errorf("encoding simplified status patch: %w", err)
-	}
-
-	if err := c.client.Status().Patch(ctx, comp, client.RawPatch(types.JSONPatchType, js)); err != nil {
+	copy := comp.DeepCopy()
```
jveski, line 144: Fixes an unrealted bug: patching /status/simplified fails if status is nil

api/v1/inputs.go
```diff
@@ -69,4 +69,9 @@ type ResourceRef struct {
 	Group   string `json:"group,omitempty"`
```
AYM1607, line 69: Unrelated, but should we start marking fields as required? I vaguely remember we trying it and something not working as expected? Maybe I'm misremembering.

internal/controllers/watch/kind.go
```diff
@@ -125,30 +126,37 @@ func (k *KindWatchController) newResourceWatchController(parent *WatchController
...
-			var exists bool
-			for _, req := range reqs {
-				if req.NamespacedName == nsn {
-					exists = true
-					break
-				}
-			}
-			if !exists {
-				reqs = append(reqs, reconcile.Request{NamespacedName: nsn})
+			req := reconcile.Request{NamespacedName: nsn}
```
chihshenghuang, line 151: The new change is updating the existing request with the name/namespace defined in comp binding, and append new request if not found? If this is correct, the logic should be:
```go
idx := slices.Index(reqs, req)
if idx >= 0 {   // >= 0
  reqs[idx] = req  // update existing request but don't need to append it
} else {
  reqs = append(reqs, req) // append request for the new request only
}
```

api/v1/inputs.go
```diff
@@ -69,4 +69,9 @@ type ResourceRef struct {
...
 	Version string `json:"version,omitempty"`
 	Kind    string `json:"kind"`
+
+	// If set, name and namespace form an "implicit binding", i.e. a ref that is bound to
+	// a specific resource without a corresponding binding on the composition resource.
+	Name      string `json:"name,omitempty"`
```
AYM1607, line 75: Worth calling out what takes precedence if an input is bount both by the synth and the composition.

internal/controllers/watch/kind.go
```diff
@@ -125,30 +126,37 @@ func (k *KindWatchController) newResourceWatchController(parent *WatchController
...
+			idx := slices.Index(reqs, req)
+			if idx > 0 {
+				reqs[idx] = req // override the ref's name/namespace if set
```
AYM1607, line 154: So the comp binding takes precedence over the synth implicit binding?

internal/controllers/watch/kind.go
jveski: Pushed a commit to fix/clarify the logic here. Implicit bindings currently take precedence over explicit so the synth will win. I _think_ this makes sense - if the synth "knows" a resource it needs, the composition shouldn't be able to say otherwise. But open to other ideas.

internal/controllers/watch/kind.go
jveski: Left a response about this on Chih-Sheng's comment.

internal/controllers/watch/kind.go
AYM1607: I rationalized it backwards. For me the Synthesizer is a default because it's the "template" and the Composition being the instantiation is more specific and should take precedence.

AYM1607: I don't think the specific behavior is important though, as long as it's consistent and called out

AYM1607 (approved):

chihshenghuang (approved):

PR #316, opened 2025-03-28
Don't validate non-defaulted resources against schema

jveski:
Some types have schemas that require fields commonly set by the defaulting functions, which don't run in Eno. Everything actually works as expected in this case except the SMD library's validation fails - we can just skip it.

### DIFF
--- a/examples/go-synthesizer/main.go
+++ b/examples/go-synthesizer/main.go
@@ -37,6 +37,7 @@ func main() {
 			Containers: []corev1.Container{{
 				Name:  "nginx",
 				Image: "nginx:latest",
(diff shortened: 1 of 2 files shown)


examples/go-synthesizer/main.go
```diff
@@ -37,6 +37,7 @@ func main() {
 			Containers: []corev1.Container{{
 				Name:  "nginx",
 				Image: "nginx:latest",
+				Ports: []corev1.ContainerPort{{ContainerPort: 80}},
```
jveski, line 40: The missing `protocol` field would cause this test to fail without the change.

chihshenghuang (approved):

AYM1607 (approved):

PR #320, opened 2025-03-31
Variation scoped SynthesisEnv

AYM1607:
This will be mostly used to enable scenarios in which a single (or a subset) variation is reconciled.


api/v1/env.go
```diff
@@ -1,9 +1,8 @@
 package v1
 
-// +kubebuilder:validation:XValidation:message="name must match [a-zA-Z_][a-zA-Z0-9_]*",rule="self.name.matches('^[a-zA-Z_][a-zA-Z0-9_]*$')"
```
AYM1607, line 3: CEL budgets are very annoying, and I don't want to limit the size of the variations array. So I'll delete this validation for now. We can consider using the eno-controller as a validating webhook in the future.

jveski (approved):
😎

PR #321, opened 2025-03-31
Better README

jveski:
- Clarify the README's elevator pitch
- Add a getting started guide
- Consolidate the synthesizer docs, re-order a few sections


jveski:
Preview: https://github.com/Azure/eno/blob/88a7eb9cec2daf0c192a7211c0b72d7305797950/README.md

README.md
```diff
@@ -1,41 +1,143 @@
 # Eno
 
-## What is Eno?
+Compose Kubernetes deployments.
 
-Eno is a new take on configuration management for Kubernetes.
+- 🎹 **Synthesize**: generate manifests dynamically in short-lived pods
+- ♻️ **Reconcile**: apply the generated configurations to one (or more!) clusters
```
AYM1607, line 6: This is technically not yet true lol, maybe local / remote?

README.md
```diff
@@ -1,41 +1,143 @@
...
+    - key: config
+      resource:
+        group: "" # core
+        version: v1
+        kind: ConfigMap
+  command:
+  - /bin/bash
+  - -c
+  - |
+    replica_count=\$(sed -n 's/.*"replicas":"\([^"]*\)".*/\1/p')
```
AYM1607, line 59: Worth adding a comment to demistify this? At least mention that this is equivalent to using `function` to read an input?

README.md
```diff
@@ -1,41 +1,143 @@
...
+    echo '{
+      "apiVersion":"config.kubernetes.io/v1",
+      "kind":"ResourceList",
+      "items":[
+        {
+          "apiVersion":"apps/v1",
+          "kind":"Deployment",
+          "metadata":{
+            "name":"my-app",
+            "namespace": "default"
```
AYM1607, line 69: Add reconcile interval? Pepople usually get suprised when their resources don't come back if deleted under Eno

README.md
jveski: The current eno-reconciler composition label selectors kind of deliver on this right?

README.md
jveski: Ooh very interesting. Unrelated to this PR, but what do you think about setting a global default in a CLI flag on the controller?

README.md
AYM1607: I was thinking about the mixed targets scenario, but this calls out "configurations" so you're right

README.md
AYM1607: I think that the semantics are clear now with the enforced min PR merged:
- Not set: Only reconciled after synthesis
- Set but under the reconciler's min: Reconciled periodically at the reconciler's minimum rate
- Set and over the reconciler's min: Reconciled periodically at the rate specified by the anno.

Totally understand if you want to leave this out though.

AYM1607 (approved):
<3

README.md
jveski: Hmm yeah that is pretty clear. Let me merge this and I'll find the right doc for this bit

PR #322, opened 2025-03-31
Support completely replacing resources instead of merging

### LINKED ISSUE #238: It should be possible to optionally replace entire resources
Maybe a particular resource should never be written to be other clients. Skip the three way merge and just replace the entire resource when it has drifted.

jveski:
Resolves #238

Some resources should only ever be written by one client. CRDs are a good example: no other clients should be able to come along and append properties to the CRD schema.


internal/resource/resource.go
```diff
@@ -301,6 +302,9 @@ func NewResource(ctx context.Context, slice *apiv1.ResourceSlice, index int) (*R
 	const disableUpdatesKey = "eno.azure.io/disable-updates"
 	res.DisableUpdates = anno[disableUpdatesKey] == "true"
 
+	const forceKey = "eno.azure.io/force"
```
chihshenghuang, line 305: any reason to use "force" to represent disable merge behavior? can we use "disable-merge" like disable-updates above? or force-recreate to make it self-explanatory

jveski: That's a great question, I'm not 100% sold on "force" - I picked it mostly because it's already a concept in helm.

That said, I suspect we may have other functionality that makes sense to include under "force" in the future, rather than having separate annotations. Support for immutable configmaps comes to mind, or playing nice with `generateName` more broadly.

tilnl: I think force is fine to use for replacing. force-recreate might be a bit more self-explanatory, but I agree it makes sense to keep just one annotation if we want to avoid having multiple similar ones.

tilnl (approved):

chihshenghuang (approved):

PR #323, opened 2025-04-01
Synth integration test helpers

jveski:
Adds `functiontest` - a small package to help test synthesizers written in Go.


pkg/functiontest/testing.go
```diff
@@ -0,0 +1,146 @@
...
+	Name        string
+	Inputs      T
+	Assertion   Assertion[T]
+	ExpectError bool
+}
+
+// Evaluate runs the synthesizer function with the provided scenarios and asserts on the outputs.
+func Evaluate[T function.Inputs](t *testing.T, synth function.SynthFunc[T], scenarios ...Scenario[T]) {
+	for _, s := range scenarios {
+		t.Run(s.Name, func(t *testing.T) {
```
bcho, line 27: can we add `t.Parallel()` by default? This is helpful for catching unexpected data race issues, although it's unlikely in this part

jveski: good idea 👍

jaiveerk (approved):
lgtm

pkg/functiontest/testing.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/Azure/eno/pkg/function"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Scenario represents a test case for a synthesizer function.
+type Scenario[T function.Inputs] struct {
+	Name        string
+	Inputs      T
+	Assertion   Assertion[T]
+	ExpectError bool
```
AYM1607, line 21: What do you think about having more expressiveness for the error expectation? Maybe an error assertion with exported defaults for "has error" and "not has error"? I'm okay with *maybe* addressing this in a follow-up?

pkg/functiontest/testing.go
```diff
@@ -0,0 +1,143 @@
...
+
+		if os.Getenv("ENO_GEN_SNAPSHOTS") != "" {
+			err = os.WriteFile(filepath.Join(dir, s.Name+".yaml"), data, 0644)
+			if err != nil {
+				t.Errorf("error while writing snapshot %q: %s", s.Name, err)
+			}
+			return
+		}
+
+		if !bytes.Equal(data, expected) {
```
AYM1607, line 123: You think it's worth having better diffing here? I also think that assuming these are being run in a git repo and relying on git's diffing is enough, but curious to hear your thoughts.

pkg/functiontest/testing_test.go
```diff
@@ -0,0 +1,77 @@
...
+		output := &corev1.Pod{}
+		output.Name = "test-pod"
+		return []client.Object{output}, nil
+	}
+
+	Evaluate(t, fn, Scenario[struct{}]{
+		Name:   "example-test",
+		Inputs: struct{}{},
+		Assertion: func(t *testing.T, scen *Scenario[struct{}], outputs []client.Object) {
+			t.Logf("asserting on %d outputs", len(outputs))
```
AYM1607, line 22: Fail if != 1?

pkg/functiontest/testing_test.go
```diff
@@ -0,0 +1,77 @@
...
+		output := &corev1.Pod{}
+		output.Name = "test-pod"
+		return []client.Object{output}, nil
+	}
+
+	Evaluate(t, fn, Scenario[struct{}]{
+		Name:   "example-test",
+		Inputs: struct{}{},
+		Assertion: AssertionChain(
+			func(t *testing.T, scen *Scenario[struct{}], outputs []client.Object) {
```
AYM1607, line 38: Should we assert that these were actually called? maybe an atomic call counter and check the count after evaluate is done running?

pkg/functiontest/testing_test.go
```diff
@@ -0,0 +1,77 @@
...
+	var inputs []map[string]any
+	var lock sync.Mutex
+	assertion := func(t *testing.T, scen *Scenario[map[string]any], outputs []client.Object) {
+		lock.Lock()
+		inputs = append(inputs, scen.Inputs)
+		lock.Unlock()
+	}
+
+	scenarios := LoadScenarios(t, "fixtures", assertion)
+	Evaluate(t, fn, scenarios...)
```
AYM1607, line 64: Check inputs?

pkg/functiontest/testing.go
jveski: Yeah that's exactly what I was thinking. Our current tooling spits out so much garbage I always just ignore it and check the git diffs anyway. But open to adding real diffs - maybe with SMD

pkg/functiontest/testing.go
jveski: What do you think about just removing `ExpectError` and letting folks write their own unit tests for error conditions? I'm thinking there aren't many error conditions for most synths anyway and they should be easy to test without getting this library involved

AYM1607: I can get behind that :)

pkg/functiontest/testing.go
AYM1607: I think relying on git tooling is the much better option for now until we can actually provide something better and more insightful.

AYM1607 (approved):
LGTM

PR #324, opened 2025-04-01
Support custom input types in pkg/function

jveski:
This change allows non-k8s types to be declared as synthesizer inputs, and bound to some corresponding k8s resource by a reusable function.

### DIFF
--- a/pkg/function/main.go
+++ b/pkg/function/main.go
@@ -10,6 +10,7 @@ import (
 
 // Inputs is satisfied by any struct that defines the inputs required by a SynthFunc.
 // Use the `eno_key` struct tag to specify the corresponding ref key for each input.
+// Each field must either be a client.Object or a custom type registered with AddCustomInputType.
 type Inputs interface{}
 
 // SynthFunc defines a synthesizer function that takes a set of inputs and returns a list of objects.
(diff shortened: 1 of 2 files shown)


pkg/function/main_test.go
```diff
@@ -38,6 +38,32 @@ func ExampleInputs() {
 	// Output: {"apiVersion":"config.kubernetes.io/v1","kind":"ResourceList","items":[{"metadata":{"creationTimestamp":null,"name":"foobar\n"},"spec":{"containers":null},"status":{}}]}
 }
 
+func ExampleAddCustomInputType() {
+	type myType struct {
+		SecretKey string
+	}
+
+	AddCustomInputType(func(in *corev1.Secret) (*myType, error) {
```
paulgmiller, line 46: are they always secrets or sometimes configmap with data?

jveski: the func is generic so anything that implements client.Object should work!

jaiveerk (approved):
lgtm

pkg/function/main.go
```diff
@@ -44,14 +45,39 @@ func main[T Inputs](fn SynthFunc[T], ir *InputReader, ow *OutputWriter) error {
 		if v.Field(i).IsNil() {
 			v.Field(i).Set(reflect.New(field.Type.Elem()))
 		}
-		err := ReadInput[client.Object](ir, tagValue, v.Field(i).Interface().(client.Object))
+		input := v.Field(i).Interface()
+
+		var obj client.Object
+		var custom bool
+		if o, ok := input.(client.Object); ok {
+			obj = o
```
AYM1607, line 53: This is here to make it so the subsequent `ReadInput` call writes to `obj` right?

pkg/function/main.go
```diff
@@ -44,14 +45,39 @@ func main[T Inputs](fn SynthFunc[T], ir *InputReader, ow *OutputWriter) error {
...
+		} else if inputType, ok := customInputSourceTypes[v.Field(i).Type().String()]; ok {
```
AYM1607, line 54: Extract this out as `key` for readability?

pkg/function/main.go
jveski: Reading through this a couple of days after writing it took me an embarrassingly long time to figure out what's going on here lol. Pushed a commit to extract this logic out to a separate func.

In any case you're right. :)

pkg/function/main_test.go
```diff
@@ -38,6 +38,32 @@ func ExampleInputs() {
...
+
+	fn := func(inputs exampleInputs) ([]client.Object, error) {
+		output := &corev1.Pod{}
+		output.Name = string(inputs.CustomInput.SecretKey)
+		return []client.Object{output}, nil
+	}
+
+	ir := newTestInputReader()
+	main(fn, ir, NewDefaultOutputWriter())
+	// Output: {"apiVersion":"config.kubernetes.io/v1","kind":"ResourceList","items":[{"metadata":{"creationTimestamp":null,"name":"foobar\n"},"spec":{"containers":null},"status":{}}]}
```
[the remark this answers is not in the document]
jveski: The example runner actually does! An underutilized feature IMO

pkg/function/main.go
```diff
@@ -113,3 +92,51 @@ func AddCustomInputType[Resource client.Object, Custom any](bind func(Resource)
...
+	// Pass through client.Object types
+	fieldVal := field.Interface()
+	if o, ok := fieldVal.(client.Object); ok {
+		i.Object = o
+		return i, nil
+	}
+
+	// Resolve custom input types back to their binding functions
+	name := field.Type().String()
+	inputType, ok := customInputSourceTypes[name]
```
AYM1607, line 119: nit: intputSourceType would make what's going on much clearer!

AYM1607 (approved):

PR #326, opened 2025-04-02
Enforce a min reconcile interval

jveski:
Don't let resources block the loop by setting a small or zero reconcile interval.


AYM1607:
You think it's worth calling this out in the docs?

jveski:
> You think it's worth calling this out in the docs?

@AYM1607 Given the quality of the docs right now, I'd say probably not 😆 . But I'll make a note to add in the future.

AYM1607 (approved):
LGTM

PR #329, opened 2025-04-08
Handle slow synthesis pod deletion

jveski:
Currently replacement synthesis pods are only created after the existing pods are fully deleted i.e. gone from the informer entirely. The intention of the design was to create a new pod as soon as the old pod is marked for deletion, so we need to also enqueue when the deletion timestamp transitions from nil to non-nil.

### DIFF
--- a/internal/controllers/synthesis/gc.go
+++ b/internal/controllers/synthesis/gc.go
@@ -36,9 +36,12 @@ func (p *podGarbageCollector) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	pod := &corev1.Pod{}
 	err := p.client.Get(ctx, req.NamespacedName, pod)
-	if err != nil || pod.DeletionTimestamp != nil {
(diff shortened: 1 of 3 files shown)


internal/controllers/synthesis/gc.go
```diff
@@ -36,10 +36,13 @@ func (p *podGarbageCollector) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	pod := &corev1.Pod{}
 	err := p.client.Get(ctx, req.NamespacedName, pod)
-	if err != nil || pod.DeletionTimestamp != nil {
-		logger.Error(err, "failed to get pod or pod is marked for deletion")
+	if err != nil {
+		logger.Error(err, "failed to get pod")
```
jveski, line 40: Unrelated, reduces log noise

chihshenghuang (approved):

PR #332, opened 2025-04-10
Move inputs helpers to separate package

jveski:
We need to call OutOfLockstep from the executor but without the full composition struct. This refactors it to take the revision slice instead of the composition, and moves both input helpers out of the API package.


chihshenghuang (approved):
🚢

PR #333, opened 2025-04-10
Guarantee synthesizer read by synthesis pod matches its own image

jveski:
It's possible for the synthesizer's image to be updated after some synthesis pods have been created, but before they've read their own synthesizer resource from apiserver. In this case it's possible that the refs and/or timeouts have changed such that the current image will never synthesize successfully. So effectively it's possible to break some in-flight syntheses by updating the image and removing refs that aren't used by the newer image.

This PR also adds a check for input revision consistency before committing the synthesis. Currently it's possible for inputs to have fallen out of lockstep between the time synthesis was dispatched and read by the executor.


chihshenghuang (approved):

jveski:
I'm going to break this up into a few smaller PRs - don't merge this one yet!

PR #336, opened 2025-04-11
Set deadline on synthesis pods

jveski:
Eno has its own timeouts for the synthesis workflow, but it still makes sense to set the built in pod deadline to cover cases where Eno's controllers are able to operate well due to high pod count or rapid churn.

This also reduces the shutdown grace period to 0 since there is no gracefully shutting down synthesizer processes.


jveski:
Hmm I'm second guessing this now. Closing until I have a clearer picture.

PR #337, opened 2025-04-15
Server-side apply

jveski:
I think it's time to accept defeat on client-side apply. 😢 

The structured-merge-diff library is great, except it doesn't (well, can't) apply default values. So the current approach works fine until an array object has a `x-kubernetes-list-map-keys` that doesn't match its `x-kubernetes-patch-merge-key`.

For example: the service port map key is protocol+port, but its merge key is only the port. Objects that set both properties can be reconciled by Eno. But setting only the port will fail, even though it would succeed with client-side three-way merge. This is because `protocol` defaults to TCP, but that default is applied in apiserver - not SMD.

We can keep hacking around this by modifying the spec, but the reasons to avoid SSA have been exhausted at this point (I'll avoid enumerating them here).


docs/reconciliation.md
```diff
@@ -5,22 +5,13 @@ Once a composition has been synthesized, the resulting resources are reconciled
 
 ## Merge Semantics
 
-Eno uses the [structured-merge-diff](https://pkg.go.dev/sigs.k8s.io/structured-merge-diff) package to merge the expected and actual states of each resource using rules defined by its schema.
+By default, Eno uses [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) to create and update resources it manages.
 
 Any property set by the synthesizer will be applied during reconciliation, even if it means overwriting changes made by another client.
 Other clients can safely populate fields _not_ managed by Eno - unmanaged fields are not modified.
+This is the standard behavior of server-side apply with `--force-conflicts=true`.
```
AYM1607, line 12: How does this impact future work on "partially managed" fileds?

docs/reconciliation.md
```diff
@@ -84,8 +75,8 @@ annotations:
 
 ### Replace
 
-Designating a resource to be replaced means that updates will not use 3-way merge,
-so any fields set by other clients will be overwritten by Eno.
+Designating a resource to be replaced means that updates will use the normaly update endpoint instead of server-side apply.
```
AYM1607, line 78: normaly -> normal?

internal/controllers/reconciliation/controller.go
```diff
@@ -50,8 +46,8 @@ type Controller struct {
 	timeout               time.Duration
 	readinessPollInterval time.Duration
 	upstreamClient        client.Client
-	discovery             *discovery.Cache
 	minReconcileInterval  time.Duration
+	disableSSA            bool
```
AYM1607, line 50: I assume we'll set this on the overlay reconciler for old cluster versions right?

internal/controllers/reconciliation/controller.go
```diff
@@ -229,40 +220,80 @@ func (c *Controller) reconcileResource(ctx context.Context, comp *apiv1.Composit
...
+			}
+			logger.V(0).Info("corrected drift in managed fields metadata")
+			return true, nil
 		}
 	}
-	if updated == nil {
-		logger.V(1).Info("skipping empty update")
+
+	// Do the actual non-dryrun update
+	reconciliationActions.WithLabelValues("apply").Inc()
```
AYM1607, line 253: After non-error?

jveski: This one was inconsistent before - all of the other operations increment before sending the request. I have no strong opinions about which way it should go, want me to move all of them after the response?

internal/controllers/reconciliation/controller.go
jveski: Yeah exactly

docs/reconciliation.md
jveski: lololol thank you

docs/reconciliation.md
jveski: Not too much I think. The reconciler still has all of the state it needs to compute the diffs - we will just need to implement that patch logic ourselves or using jsonpatch rather than tieing into the fieldpath bits of structured-merge-diff.

internal/controllers/reconciliation/controller.go
AYM1607: Ah I didn't realize this was the only inconsistent one, I'm good with it then :)

AYM1607 (approved):

PR #343, opened 2025-04-18
Use Docker cache mount for go mods in smoke tests

jveski:
Speeds up the smoke tests pretty significantly by avoiding doing `go mod download` from all of the image builds.


chihshenghuang (approved):
🚢

PR #359, opened 2025-04-28
Remove deprecated fields

jveski:
Works as advertised


api/v1/types.go
```diff
@@ -7,7 +7,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/scheme"
 )
 
-//go:generate controller-gen object crd rbac:roleName=resourceprovider paths=./...
+//go:generate go run sigs.k8s.io/controller-tools/cmd/controller-gen object crd rbac:roleName=resourceprovider paths=./...
```
AYM1607, line 10: Nice! No need to have the binary installed

AYM1607 (approved):

PR #361, opened 2025-04-29
Add example for CRD+CR

jveski:
- Adds example to prove that CRDs and CRs can be created in the same composition
- Adds helper function to read manifests from disk (in synthesizers)
- Output writer now attempts to resolve the GVK of a resource from its struct if the value isn't specified


chihshenghuang (approved):
🚢

PR #362, opened 2025-04-30
Unblock composition deletion when object's kind doesn't exist

jveski:
Deleting a resource that doesn't have a registered kind doesn't return a 404 so we need special logic to handle this case.

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -122,7 +122,7 @@ func (c *Controller) Reconcile(ctx context.Context, req resource.Request) (ctrl.
 
 	// Fetch the current resource
 	current, err := c.getCurrent(ctx, resource)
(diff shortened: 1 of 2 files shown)


internal/controllers/reconciliation/controller.go
```diff
@@ -349,3 +349,10 @@ func isErrMissingNS(err error) bool {
 	}
 	return strings.Contains(err.Error(), "an empty namespace may not be set")
 }
+
+func isErrNoKindMatch(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "no matches for kind")
```
AYM1607, line 357: There's no exported helper from k8s errors or api?

jveski: Nope, and `errors.ReasonForError` returns a zero value 🤷

AYM1607 (approved):

PR #364, opened 2025-05-02
Fix flaky test

jveski:
One of the SSA field ownership tests is flaky due to a race condition that _technically_ applies to actual use although it's unlikely to show up outside of this particular test.

The problem is that we only remediate missing `managedFields` metadata when a previous expected state exists. So if field ownership is removed before the composition's second synthesis it will be re-introduced __by sending another SSA patch request__ not using our out-of-band field ownership remediation (hack). The problem is that apiserver will consider existing fields to be owned by `before-first-apply` in addition to the eno field manager. So if the fields are removed at any point in the future, they will be pruned from Eno's ownership but not actually removed from the resource because `before-first-apply` still holds a reference to them.

The fix is to populate any missing field ownership metadata proactively in this case instead of letting apiserver handle it.


AYM1607 (approved):
:shipit:

PR #365, opened 2025-05-02
Remove noisy 404 log

jveski:
We expect and tolerate 404s in this case but still log them as if they were real/unexpected errors.


AYM1607 (approved):
:ship:

PR #366, opened 2025-05-04
fix(helmshim): skip nil helm object

bcho:
Helm renderer can generate empty yaml file like this:

```yaml
---
# source: a.yaml
# comment
---
# source: b.yaml
apiVersion: v1
kind: ConfigMap
# ... object data
```

In this case, the section of a.yaml will be decoded as an empty `unstructured.Unstructred` object. However, object without GVK is not supported in eno and it will be rejected in the output writer: https://github.com/Azure/eno/blob/07e348b023dc93833fb509ff6f8aaa2b014e173a/pkg/function/outputs.go#L56-L63

But in current helmshim implementation, the return value of output writer `Write` is ignored, hence it suppressed the underlying issue.

This pull request addressed this issue by:

1. handle error from output writer `Write` method
2. skip helm objects that rendered as empty `unstructured.Unstructured` object by checking on the JSON representation.


pkg/helmshim/helm.go
```diff
@@ -32,6 +33,19 @@ func MustRenderChart(opts ...RenderOption) {
...
+// This is used when then input helm chart includes an empty target (for example: empty yaml file with comments).
+func isNullObject(o *unstructured.Unstructured) bool {
+	if o == nil {
+		return true
+	}
+	b, err := json.Marshal(o)
+	if err != nil {
+		return false
+	}
+	return string(b) == "null"
```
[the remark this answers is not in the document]
bcho: Would still delegate to the JSON marshaler used by unstructured scheme to provide the correct value

pkg/helmshim/helm.go
```diff
@@ -32,6 +33,19 @@ func MustRenderChart(opts ...RenderOption) {
 	}
 }
 
+// isNullObject checks if the given unstructured object is equivalent to an empty K8S object.
+// This is used when then input helm chart includes an empty target (for example: empty yaml file with comments).
+func isNullObject(o *unstructured.Unstructured) bool {
+	if o == nil {
+		return true
+	}
+	b, err := json.Marshal(o)
```
jveski, line 42: I'd prefer to avoid the overhead of json encoding here. Would the test pass with `o == nil || o.Object == nil`?

pkg/helmshim/helm_test.go
```diff
@@ -7,8 +7,45 @@ import (
...
+		},
+		{
+			name: "empty object",
+			o:    &unstructured.Unstructured{},
+			want: true,
+		},
+		{
+			name: "non-empty object",
+			o: &unstructured.Unstructured{
+				Object: map[string]any{
```
jveski, line 32: Needs another test case to cover `o.Object = map[string]any{}`

pkg/helmshim/helm.go
bcho: yes and i think these two checks should handle 99% of the cases. But becuase the unstructured.Unstructured is using a customized JSON marshaler: https://github.com/kubernetes/apimachinery/blob/955939ffb8198b47fe11e4e53121d529ee6f946b/pkg/apis/meta/v1/unstructured/unstructured.go#L124 therefore i want to avoid some edge cases that we are not aware.

pkg/helmshim/helm_test.go
bcho: technically helm should not return value like this because it's also checking on gvk if the yaml file is not empty. Added a validation logic on this too, but this will result in we have to use json.Marshal to generate the encoded result.

pkg/helmshim/helm.go
jveski: agreed - best to hedge against changes to the marshaler since we will hit the branch on line 42 most of the time anyway

jveski (approved):
🚢

PR #387, opened 2025-05-21
Improve k8s version matrix test tooling

jveski:
kubebuilder removed support for older k8s versions so we need to roll an alternative for matrix testing.

- Downloads artifacts directly from the k8s distribution site
- Builds the test matrix from metadata published to the same site
- Only builds the tests once!


hack/build-k8s-matrix.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+start_minor=22 # TODO: lower this to ~10
```
chihshenghuang, line 5: this is used to support sever-side apply starting from 1.22?

chihshenghuang (approved):

hack/build-k8s-matrix.sh
jveski: It's just an arbitrary version that is currently passing. I'll send another PR to fix older versions and update this value.

PR #397, opened 2025-05-27
Fix ownership backfill race

jveski:
Currently Eno will attempt to backfill missing managedFields metadata by comparing the current value against a dryrun of applying the previous and next versions of the resource. This works for the previous state since the goal is to avoid removing properties without ensuring that they will be pruned properly. But it makes _zero_ sense for the next state - it may contain properties that haven't been applied yet.

Adding the managedFields early isn't inherently a bad idea, except that apiserver asynchronously prunes them. So _sometimes_ the values will be removed faster than Eno's retry interval, which effectively deadlocks reconciliation.

There isn't a good way to integration test this behavior, but I've run 10+ iterations of the integration tests and this fixes the most common flake.

### DIFF
--- a/internal/resource/resource.go
+++ b/internal/resource/resource.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"hash/fnv"
 	"maps"
-	"reflect"
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


internal/resource/resource.go
```diff
@@ -314,7 +313,7 @@ func CompareEnoManagedFields(a, b []metav1.ManagedFieldsEntry) bool {
 	if ai == -1 || ab == -1 {
 		return false
 	}
-	return reflect.DeepEqual(a[ai], b[ab])
+	return equality.Semantic.DeepEqual(a[ai], b[ab])
```
jveski, line 316: Unrelated: just seems like a good idea since the managedFields struct contains timestamps

AYM1607 (approved):

PR #400, opened 2025-05-28
Fix TestInputSynthesizerOrdering flake

jveski:
I don't think there's actually anything wrong with the code being tested - updating the test to fix the flake.


AYM1607:
Well this is a weird one lol

PR #401, opened 2025-05-29
Support partial reconciliation

jveski:
Some synthesizers may need to allow other clients to modify fields they set. Rather than feedback the current state of every resource into the synthesizer (expensive, confusing), let's support a simple jsonpatch-style interface.

### DIFF
--- a/docs/reconciliation.md
+++ b/docs/reconciliation.md
@@ -91,16 +91,23 @@ annotations:
     ]
 ```
 
+This is commonly used to make a subset of properties managed by Eno optional i.e. allow other clients to override them.
+For example:
+
+```json
+{ "path": "self.data.foo", "value": "default value", "condition": "!has(self.data.foo)" }
+```
+
 #### Path Expression Syntax
 
-Overrides use a jsonpath-esque syntax to target properties.
+Overrides use a CEL-like syntax to reference properties.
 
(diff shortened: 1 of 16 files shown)


docs/reconciliation.md
```diff
@@ -73,6 +73,36 @@ annotations:
...
+
+This allows Eno synthesizers to specify basic runtime behavior without requiring resynthesis.
+
+> Overrides are applied during reconciliation, so in most cases they should be used alongside `eno.azure.io/reconcile-interval`.
+
+```yaml
+annotations:
+  eno.azure.io/overrides: |
+    [
+      { "path": "/data/foo", "value": "new conditional value", "condition": "self.data.bar == 'baz'" }
```
chihshenghuang, line 90: is that possible to use self in value field? like {"path": "/data/foo", "value": "self.spec.containers[0].resources.limits.cpu"}

docs/reconciliation.md
```diff
@@ -73,6 +73,36 @@ annotations:
   eno.azure.io/disable-updates: "true"
 ```
 
+### Overrides
```
AYM1607, line 76: By far the most common use case for this will be supporting external controller ownership of a certain subset of fields. Might be worth exploring that here?

chihshenghuang (approved):
🚀

PR #414, opened 2025-06-27
Allow users to write Synthesizer with KCL

wonderyl:
As title


pkg/kclshim/example_input.json
```diff
@@ -0,0 +1,17 @@
+{
```
jveski, line 1: Mind moving all of the test data into `pkg/kclshim/fixtures`? It's a pattern used by other packages in eno

pkg/kclshim/kcl.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		rl.Results = append(rl.Results, &krmv1.Result{
+			Message:  fmt.Sprintf("error marshaling error response: %v", err),
+			Severity: krmv1.ResultSeverityError,
+		})
+	}
+	fmt.Print(string(bytes))
+}
+
+func Synthesize(workingDir string) {
```
jveski, line 37: Is it feasible to use `kcl run` directly in a synthesizer container and skip the go process entirely? I'm not familiar enough with KCL to know if there are any gaps there but I like the idea

pkg/kclshim/example_synthesizer/main.k
```diff
@@ -0,0 +1,4 @@
+import pkg
```
jveski, line 1: Should the `pkg` kcl package be renamed `eno`?

pkg/kclshim/example_synthesizer/pkg/resource_list.k
```diff
@@ -0,0 +1,25 @@
...
+
+schema ResultResourceRef:
+    apiVersion: str
+    kind: str
+    name: str
+    namespace?: str
+
+schema Result:
+    message: str
+    file?: ResultFile
```
jveski, line 15: Eno doesn't support result files fwiw

pkg/kclshim/example_synthesizer/pkg/resource_list.k
```diff
@@ -0,0 +1,25 @@
+# TODO: Move this file to a KCL module.
```
jveski, line 1: I agree with this TODO 🙃

pkg/kclshim/example_synthesizer/pkg/synthesizer.k
```diff
@@ -0,0 +1,10 @@
+Synthesize: (ResourceList) -> ResourceList = lambda input: ResourceList -> ResourceList {
+    mcr_base = [item.data.mcr_base for item in input.items if item.metadata.name == "some-config"][0]
```
jveski, line 2: `image_base` instead of `mcr_base`?

pkg/kclshim/kcl.go
wonderyl: KCL can't read stdin directly. The KCL owners created a separate project https://github.com/kcl-lang/krm-kcl, that wraps it with golang. I didn't use it for 2 reasons. 
1. It assumes the stdin is in yaml format, while we use json.
2. It expects a "FunctionConfig" from input as well as a ResourceList.

jveski, line 37: Let's get rid of the README and use godoc comments instead - on this function and maybe on the top-level package. Doesn't need to be too detailed - just something to explain how to get started with this library

go.mod
```diff
@@ -72,13 +80,16 @@ require (
...
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250528174236-200df99c418a // indirect
+	google.golang.org/grpc v1.72.0 // indirect
 	google.golang.org/protobuf v1.36.6 // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/apiextensions-apiserver v0.33.1 // indirect
 	k8s.io/klog/v2 v2.130.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20250318190949-c8a335a9a2ff // indirect
+	kcl-lang.io/kcl-go v0.11.2 // indirect
```
jveski, line 91: I'd prefer to keep these dependencies out of the main module since they aren't used by Eno itself and might cause conflicts or just dependency bloat when folks import the synthesizer SDK (pkg/function).

The project uses go workspaces so you should be able to initialize a new go module in pkg/kclshim and add it to the workspace. Helmshim already does this today

jveski (approved):

PR #415, opened 2025-06-27
tests but they broke

paulgmiller:


pkg/functiontest/synthlint.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// Extract ref keys
+	var refKeys []string
+	for _, ref := range synthesizer.Spec.Refs {
+		if ref.Key != "" {
+			refKeys = append(refKeys, ref.Key)
+		}
+	}
+
+	if len(refKeys) == 0 {
```
jveski, line 102: same question as on the struct func - haven't thought about it much but I think this should allow an empty list

pkg/functiontest/synthlint_test.go
```diff
@@ -0,0 +1,247 @@
...
+	// Test with subset of refs - should pass in relaxed mode
+	inputs := TestInputsPartial{} // only has database and cache
+
+	// This should not fail in relaxed mode since eno_keys are a subset of refs
+	InputsMatchSynthesizerRefs(t, inputs, synthesizerPath, KeyMatchRelaxed)
+}
+
+func TestInputsMatchSynthesizerRefs_RelaxedMode_FailureMissingRef(t *testing.T) {
+	// Create synthesizer.yaml with fewer refs than eno_keys
+	synthesizerPath := createTempSynthesizer(t, []string{"database", "cache"})
```
jveski, line 97: personally I'd prefer a couple of test fixtures instead of this helper. Nice to read the fixtures directly when debugging failing tests

pkg/functiontest/synthlint.go
```diff
@@ -0,0 +1,106 @@
...
+	for i := range structType.NumField() {
+		field := structType.Field(i)
+
+		// Check if field has eno_key tag
+		if enoKey, ok := field.Tag.Lookup("eno_key"); ok {
+			enoKeys = append(enoKeys, enoKey)
+		}
+	}
+
+	if len(enoKeys) == 0 {
```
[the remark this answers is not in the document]
paulgmiller: No but it should be assert. In a previous formuattion treating it as an error made for simpler code but no longe true. Moved to InputsMatchSynthesizerRefs as a requrie notempty.

pkg/functiontest/synthlint.go
```diff
@@ -0,0 +1,106 @@
...
+	"os"
+	"reflect"
+
+	enov1 "github.com/Azure/eno/api/v1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+// Define a new type for your matching mode instead of using a bare bool or int
```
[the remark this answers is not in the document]
paulgmiller: obliterated.

pkg/functiontest/synthlint_test.go
paulgmiller: By fixtures you mean checked in files right? Thats interestin y instict was to go the opposite way make InputsMatchSynthesizerRefs take a reader and never goto disk at all.

paulgmiller: Maybe I am just burned from all the file regeneration in ccp/control-plane-core. 

AI seems happy to generate thos yamls and it does make the code shorter. Pushing that soon

jveski (approved):

PR #419, opened 2025-07-02
Optionally preserve resources during composition deletion

jveski:
We need a way to avoid deadlocks in obscure cases where early stages of composition deletion might break the deletion of later resources.

### DIFF
--- a/internal/controllers/reconciliation/edgecase_test.go
+++ b/internal/controllers/reconciliation/edgecase_test.go
@@ -12,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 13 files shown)


internal/controllers/reconciliation/edgecase_test.go
```diff
@@ -570,3 +570,48 @@ func TestSynthesisErrorResult(t *testing.T) {
...
+				},
+			},
+		}}
+		return output, nil
+	})
+
+	// Test subject
+	setupTestSubject(t, mgr)
+	mgr.Start(t)
+	_, comp := writeGenericComposition(t, upstream)
```
AYM1607, line 598: The composition does not have the label, is this supposed to pass?

internal/controllers/resourceslice/slicecleanup_test.go
```diff
@@ -53,6 +54,39 @@ func TestSliceCleanupSliceReferences(t *testing.T) {
...
+	require.NoError(t, cli.Delete(ctx, slice))
+	req := reconcile.Request{NamespacedName: client.ObjectKeyFromObject(slice)}
+
+	comp.Status.CurrentSynthesis = &apiv1.Synthesis{ResourceSlices: []*apiv1.ResourceSliceRef{{Name: slice.Name}}}
+	require.NoError(t, cli.Status().Update(ctx, comp))
+
+	_, err := c.Reconcile(ctx, req)
+	require.NoError(t, err)
+
+	_, err = c.Reconcile(ctx, req)
```
chihshenghuang, line 84: why reconcile twice? or this is typo?

jveski: Once to delete the resource slice, once to remove its finalizer :)

PR #423, opened 2025-07-02
Orphan managed resources during symphony deletion

jveski:
# (kind of a) breaking change!

Cascading deletion caused by the deletion of a symphony will no longer wait for managed resources to be cleaned up. For the sake of maintaining deterministic behavior, they won't be deleted at all in this case.

The idea is to protect against deadlock states introduced by impossible cleanup steps - either imposed by the synthesizer's logic (structure of the resources themselves), availability of apiserver, or availability of the eno-reconciler process.


chihshenghuang (approved):
definitely a simple way to fix the deadlock issue 🚀

PR #425, opened 2025-07-03
failed closed on three eno annotations if they parse bad. add tests b…

paulgmiller:
…ut still some failures


internal/resource/resource.go
```diff
@@ -244,7 +244,7 @@ func NewResource(ctx context.Context, slice *apiv1.ResourceSlice, index int) (*R
 	if str, ok := anno[reconcileIntervalKey]; ok {
 		reconcileInterval, err := time.ParseDuration(str)
 		if anno[reconcileIntervalKey] != "" && err != nil {
-			logger.V(0).Info("invalid reconcile interval - ignoring")
+			return nil, fmt.Errorf("invalid reconcile interval, %s", str)
```
jveski, line 247: I _think_ it's safe to make these terminal errors but we should add an integration test in the reconciliation controller to prove that producing invalid resources doesn't cause a deadlock state e.g. synthesis is blocked because the previous synthesis is invalid

PR #434, opened 2025-07-10
Support overrides in patches

jveski:
Eno `Patch` meta-resources now support the `eno.azure.io/overrides` annotation.

This change also improves the patch mechanism such that it better resembles the normal non-patch workflow. Eno now relies on apiserver to apply patches - no more applying patches in memory and comparing against the current resource.


chihshenghuang (approved):
🚢

Issue #443, opened 2025-07-18
Field names with hyphens are not usable in eno.azure.io/overrides


SRodi:
## Description
Eno reconciler currently does not support annotations referencing fields with hyphens in their names within either the path or condition fields of an override rule. Both fields rely on a dot notation syntax that only works with standard identifier names—i.e., fields made up of letters, digits, and underscores. When a key contains a hyphen (e.g. `foo-bar`), it cannot be accessed in dot notation (`self.data.foo-bar`), and ENO does not support bracket notation (`self.data['foo-bar']`) as an alternative. This makes it impossible to reference such keys directly in override expressions.

## Issue
As a result, users are unable to create conditions or target values based on hyphenated keys in their input objects, even though these are valid YAML keys.

## Workaround
The only current workaround is to avoid using hyphens in key names altogether or to introduce a duplicate alias key that uses underscores. This limitation reduces the flexibility of `eno.azure.io/overrides`, particularly when working with third-party ConfigMaps or Helm charts that do use hyphens.

## Request
Support for bracket notation or quoted keys in override expressions would solve this and improve ENO’s robustness in real-world configurations.

## Expected behavior
This is  currently not working:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: example-config
  annotations:
    eno.azure.io/overrides: |
      [
        { "path": "self.data['foo-bar']", "value": null, "condition": "has(self.data['foo-bar']) && !pathManagedByEno" }
      ]
data:
  foo-bar: "test"
```

## Environment
```txt
Kubectl Client Version: v1.31.1
Kustomize Version: v5.4.2
Kubernetes Server Version: v1.32.5
eno Version: v0.1.25
```

PR #444, opened 2025-07-18
feat: support hyphenated field names in overrides with bracket notation

SRodi:
## Description
* Parser enhancement: Added custom `lexer` with support for both single and double-quoted strings in bracket notation for hyphenated field names
* Mutation logic fix: Implemented `unquoteKey()` function to handle both single and double quotes, replacing simple string slicing with quote handling
* Docs: Added bracket notation syntax to `reconciliation.md` path expression documentation
* Integration tests: Added 3 new integration tests in `overrides_test.go`
  - Basic hyphenated field override with both quote styles
  - Conditional overrides using hyphenated fields in both path and condition
  - Polling scenario with hyphenated fields and reconcile intervals
* Unit tests: Enhanced `mutation_test.go` with additional test cases for hyphenated field names using bracket notation
* Core functionality: Enabled support for `field["hyphen-key"]` and `field['hyphen-key']` syntax while maintaining backward compatibility with existing dot notation

## Related Issue
#443

### DIFF
--- a/internal/controllers/reconciliation/overrides_test.go
+++ b/internal/controllers/reconciliation/overrides_test.go
@@ -349,3 +349,169 @@ func TestOverrideManagedFields(t *testing.T) {
 		return cm.Data["foo"] == "bar"
 	})
 }
+
+func TestOverrideHyphenatedFieldNames(t *testing.T) {
+	ctx := testutil.NewContext(t)
+	mgr := testutil.NewManager(t)
+	upstream := mgr.GetClient()
+
+	registerControllers(t, mgr)
+	testutil.WithFakeExecutor(t, mgr, func(ctx context.Context, s *apiv1.Synthesizer, input *krmv1.ResourceList) (*krmv1.ResourceList, error) {
+		output := &krmv1.ResourceList{}
+		output.Items = []*unstructured.Unstructured{{
+			Object: map[string]any{
+				"apiVersion": "v1",
+				"kind":       "ConfigMap",
+				"metadata": map[string]any{
+					"name":      "test-obj",
+					"namespace": "default",
+					"annotations": map[string]any{
+						"eno.azure.io/overrides": `[
+							{"path": "self.data[\"hyphen-key\"]", "value": "double-quote-value"},
+							{"path": "self.data['single-hyphen-key']", "value": "single-quote-value"},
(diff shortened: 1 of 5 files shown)


internal/controllers/reconciliation/overrides_test.go
```diff
@@ -349,3 +349,169 @@ func TestOverrideManagedFields(t *testing.T) {
 		return cm.Data["foo"] == "bar"
 	})
 }
+
+func TestOverrideHyphenatedFieldNames(t *testing.T) {
```
jveski, line 353: Let's either merge all of the new tests into a single integration test, or remove them all and make sure we have enough test coverage in the parser itself.

The reason being that these tests are pretty slow and we already have pretty good coverage of the override codepath

internal/resource/mutation/mutation.go
```diff
@@ -69,6 +69,23 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
 	return Apply(o.Path, mutated.Object, o.Value)
 }
 
+// unquoteKey removes quotes from a key string, handling both single and double quotes
+func unquoteKey(key string) string {
+	if len(key) >= 2 {
```
jveski, line 74: nit: early return instead

internal/resource/mutation/mutation.go
```diff
@@ -69,6 +69,23 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
...
+		if (key[0] == '"' && key[len(key)-1] == '"') || (key[0] == '\'' && key[len(key)-1] == '\'') {
```
jveski, line 75: Maybe a regex would be simpler? And it could cover both the single and double quote case

internal/resource/mutation/mutation_test.go
```diff
@@ -345,7 +361,8 @@ func TestInvalidPathInJson(t *testing.T) {
 	assert.Error(t, err)
 	assert.Len(t, ops, 1)
 	for _, op := range ops {
-		assert.NoError(t, Apply(op.Path, map[string]any{}, op.Value))
+		// The main goal is to ensure no panic occurs, errors are acceptable
+		_ = Apply(op.Path, map[string]any{}, op.Value)
```
jveski, line 365: nice 👍

jveski (approved):

PR #450, opened 2025-07-28
Fix overrides when applied to container resource fields

jveski:
- Single quotes are handled correctly by SMD expressions (e.g. field ownership checks) but not when actually applying overrides (leftover of #444)
- The override `Apply` func doesn't prune maps that are empty after overrides, which apiserver expects

This PR also does a bit of refactoring to both `Apply` and `unquoteKey` - hopefully the diff isn't too difficult to read.

### DIFF
--- a/internal/controllers/reconciliation/overrides_test.go
+++ b/internal/controllers/reconciliation/overrides_test.go
@@ -8,8 +8,11 @@ import (
 	apiv1 "github.com/Azure/eno/api/v1"
 	"github.com/Azure/eno/internal/testutil"
 	krmv1 "github.com/Azure/eno/pkg/krm/functions/api/v1"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


internal/controllers/reconciliation/overrides_test.go
```diff
@@ -441,3 +444,98 @@ func TestOverrideHyphenatedFieldNames(t *testing.T) {
...
+	// Simulate another client updating the resources value
+	err := retry.RetryOnConflict(testutil.Backoff, func() error {
+		err := mgr.DownstreamClient.Get(ctx, client.ObjectKeyFromObject(deploy), deploy)
+		if err != nil {
+			return err
+		}
+		deploy.Spec.Template.Spec.Containers[0].Resources.Limits["cpu"] = resource.MustParse("20m")
+		return mgr.DownstreamClient.Update(ctx, deploy)
+	})
+	require.NoError(t, err)
```
AYM1607, line 534: Nit: Add a step to fetch the object with its managed fields and verify that it's no longer managed by Eno.

internal/resource/mutation/mutation.go
```diff
@@ -74,12 +77,20 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
...
-	if matches := quotedStringRegex.FindStringSubmatch(key); matches != nil {
-		if matches[1] == matches[3] {
-			return matches[2]
-		}
+	matches := quotedStringRegex.FindStringSubmatch(key)
+	if matches == nil || matches[1] != matches[3] {
+		return key
+	}
+
+	content := matches[2]
```
AYM1607, line 85: So essentially this is handling strings of the following forms:
- `"hello\"there"`
- `'hello\'there'`

right?

internal/resource/mutation/mutation.go
```diff
@@ -103,44 +114,47 @@ func apply(path *PathExpr, startIndex int, obj any, value any) error {
 	state := obj
 
 	for i, section := range path.ast.Sections[startIndex:] {
+		isLastSection := startIndex+i == len(path.ast.Sections)-1
+
 		// Map field indexing
-		if section.Field != nil {
+		if section.Field != nil || (section.Index != nil && section.Index.Key != nil) {
```
AYM1607, line 120: Which scenarios do these 2 conditionals match?

internal/resource/mutation/mutation.go
```diff
@@ -103,44 +114,47 @@ func apply(path *PathExpr, startIndex int, obj any, value any) error {
...
-			continue // should be impossible
-		}
-
-		// Alternative map field indexing
-		if key := section.Index.Key; key != nil {
-			m, ok := state.(map[string]any)
-			if !ok {
-				continue
+			var keyStr string
+			if section.Field != nil {
```
AYM1607, line 127: When is this pre-populated vs not?

internal/resource/mutation/mutation.go
```diff
@@ -103,44 +114,47 @@ func apply(path *PathExpr, startIndex int, obj any, value any) error {
...
+				keyStr = *section.Field
+			} else {
+				keyStr = unquoteKey(*section.Index.Key)
 			}
-			keyStr := unquoteKey(*key)
-			if startIndex+i == len(path.ast.Sections)-1 {
+
+			if isLastSection {
```
AYM1607, line 133: Why only when it's the last section?

internal/resource/mutation/mutation.go
```diff
@@ -103,44 +114,47 @@ func apply(path *PathExpr, startIndex int, obj any, value any) error {
...
 				if value == nil {
 					delete(m, keyStr)
 				} else {
 					m[keyStr] = value
 				}
 				return nil
 			}
-			state = m[keyStr]
+
+			nextState := m[keyStr]
```
AYM1607, line 142: What does state refer to here?

internal/resource/mutation/mutation_test.go
```diff
@@ -230,6 +229,72 @@ func TestApply(t *testing.T) {
...
+						},
+					},
+					"template": map[string]any{
+						"metadata": map[string]any{
+							"labels": map[string]any{
+								"foo": "bar",
+							},
+						},
+						"spec": map[string]any{
+							"containers": []any{
```
AYM1607, line 250: Worth testing multiple containers that match name='foo'? Or is that not a supported scenario?

internal/resource/mutation/mutation.go
jveski: `section.Field != nil` is a field like `parent.field` and ` (section.Index != nil && section.Index.Key != nil) ` is a field like `parent["field"]`

jveski: They were separate branches before with nearly identical code - refactored to flatten them into this branch

internal/resource/mutation/mutation.go
jveski: I think the other comment about the logic addresses this?

internal/resource/mutation/mutation_test.go
jveski: Overrides will work for fields that don't use `ownedByEno`. But the ownership check only works when matching against the key of the map slice since that's the key used in managedFields (docs coming soon on that lol). Since map slice keys are unique it isn't possible to have multiple matches

internal/controllers/reconciliation/overrides_test.go
jveski: I'll make a note to go back and fix this + improve the override tests more broadly. But I don't have time today and would like to get this merged to unblock folks 😬

AYM1607 (approved):

PR #458, opened 2025-07-31
replace-when annotation

AYM1607:
Allow replacing a resource conditionally based on its current state.


docs/reconciliation/README.md
```diff
@@ -16,6 +16,17 @@ Exceptions:
 
 - The eno-reconciler process can fall back to client-side three-way merge patch by setting `--disable-ssa`
 - Merge can be disabled for a resource by setting the `eno.azure.io/replace: "true"` annotation (a full `update` request will be used instead of a `patch`)
+- Merge can be disabled conditionally by setting the `eno.azure.io/replace-when` annotation with a CEL expression, it takes precedence over the non-conditional replace:
```
AYM1607, line 19: Precedence is arguable, maybe we should fail closed if they're both present.

PR #461, opened 2025-08-01
Add tests for overriding eno.azure.io/replace

jveski:
Proves that resources can be conditionally replaced by Eno.

### DIFF
--- a/internal/controllers/reconciliation/overrides_test.go
+++ b/internal/controllers/reconciliation/overrides_test.go
@@ -539,3 +539,97 @@ func TestOverrideContainerResources(t *testing.T) {
 	assert.True(t, deploy.Spec.Template.Spec.Containers[0].Resources.Limits["cpu"].Equal(resource.MustParse("20m")))
 	assert.True(t, deploy.Spec.Template.Spec.Containers[0].Resources.Requests["cpu"].Equal(resource.MustParse("5m")))
 }
+
+func TestOverrideReplaceAnnotation(t *testing.T) {
+	ctx := testutil.NewContext(t)
+	mgr := testutil.NewManager(t)
+	upstream := mgr.GetClient()
+
+	registerControllers(t, mgr)
+	testutil.WithFakeExecutor(t, mgr, func(ctx context.Context, s *apiv1.Synthesizer, input *krmv1.ResourceList) (*krmv1.ResourceList, error) {
+		output := &krmv1.ResourceList{}
+		output.Items = []*unstructured.Unstructured{{
+			Object: map[string]any{
+				"apiVersion": "v1",
+				"kind":       "ConfigMap",
+				"metadata": map[string]any{
+					"name":      "test-obj",
+					"namespace": "default",
+					"annotations": map[string]any{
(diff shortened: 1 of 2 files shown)


internal/controllers/reconciliation/overrides_test.go
```diff
@@ -539,3 +539,97 @@ func TestOverrideContainerResources(t *testing.T) {
 	assert.True(t, deploy.Spec.Template.Spec.Containers[0].Resources.Limits["cpu"].Equal(resource.MustParse("20m")))
 	assert.True(t, deploy.Spec.Template.Spec.Containers[0].Resources.Requests["cpu"].Equal(resource.MustParse("5m")))
 }
+
+func TestOverrideReplaceAnnotation(t *testing.T) {
```
AYM1607, line 543: Cool, so this essentially creates a resource through reconcile, make an update from an external client that causes the resource to be replaced, Eno replaces the resource thus removing the external client changes (including the one that caused the replace in the first place), make an external update again that is now not replaced?

internal/controllers/reconciliation/overrides_test.go
```diff
@@ -539,3 +539,97 @@ func TestOverrideContainerResources(t *testing.T) {
...
+		output := &krmv1.ResourceList{}
+		output.Items = []*unstructured.Unstructured{{
+			Object: map[string]any{
+				"apiVersion": "v1",
+				"kind":       "ConfigMap",
+				"metadata": map[string]any{
+					"name":      "test-obj",
+					"namespace": "default",
+					"annotations": map[string]any{
+						"eno.azure.io/overrides": `[{"path": "self.metadata.annotations['eno.azure.io/replace']", "value": "true", "condition": "self.metadata.annotations['should-replace'].startsWith('tru')"}]`,
```
AYM1607, line 559: Nit: Use double quote strings? We don't have backtick strings in yaml

internal/resource/resource_test.go
```diff
@@ -81,6 +81,59 @@ var newResourceTests = []struct {
...
+		Name: "conditional replace override, positive",
+		Manifest: `{
+			"apiVersion": "apps/v1",
+			"kind": "Deployment",
+			"metadata": {
+				"name": "foo",
+				"namespace": "bar",
+				"annotations": {
+					"trigger-replace": "yespls",
+					"eno.azure.io/overrides": "[{\"path\":\"self.metadata.annotations[\\\"eno.azure.io/replace\\\"]\", \"value\":\"true\", \"condition\": \"self.metadata.annotations[\\\"trigger-replace\\\"] == 'yespls'\"}]"
```
AYM1607, line 111: test startsWith? I tried and I guess couldn't get the escaping right

internal/controllers/reconciliation/overrides_test.go
jveski: Yep! 💯

internal/resource/resource_test.go
jveski: The integration test uses it and I did some manual tests too. Think we should add some to `internal/cel`?

internal/controllers/reconciliation/overrides_test.go
AYM1607: works for me

AYM1607 (approved):

PR #463, opened 2025-08-01
Add helper for comparing resource quantities

jveski:
It's hard to compare resource quantity strings using cel, but we can easily expose the k8s library to the cel runtime.

### DIFF
--- a/internal/cel/cel_test.go
+++ b/internal/cel/cel_test.go
@@ -44,3 +44,223 @@ func TestEvalFloatTypeCoersion(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, true, val.Value())
 }
+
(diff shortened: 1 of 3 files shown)


internal/cel/cel_test.go
```diff
@@ -44,3 +44,223 @@ func TestEvalFloatTypeCoersion(t *testing.T) {
...
+			expected: 0,
+		},
+		{
+			name:     "Ti vs Gi",
+			left:     "1Ti",
+			right:    "1024Gi",
+			expected: 0,
+		},
+
+		// Scientific notation (using valid K8s format)
```
AYM1607, line 162: Nit, this is not scientific notation and the test is repeated.

AYM1607 (approved):

PR #471, opened 2025-08-06
Symphony 2pc annotation protocol

jveski:
Removes annotations from all compositions before adding/updating them (see updated docs for more).

This is a building block to help "migrate" resources between synthesizers.

### DIFF
--- a/docs/symphony/README.md
+++ b/docs/symphony/README.md
@@ -1,6 +1,6 @@
 # Symphony
 
-> ⚠️ This is an advanced Eno concept, most use-cases should not use it.
+> ⚠️ This is an advanced Eno concept, most use-cases can be addressed by managing individual compositions.
 
 Symphonies are higher-order units of configuration that involve multiple synthesizers.
 


api/v1/symphony.go
```diff
@@ -31,14 +30,6 @@ type SymphonySpec struct {
 	// Synthesizer refs must be unique across variations.
 	// Removing a variation will cause the composition to be deleted!
 	Variations []Variation `json:"variations,omitempty"`
-
-	// Bindings are inherited by all compositions managed by this symphony.
```
AYM1607, line 35: We still haven't gotten rid of these in AKS, just the shared SynthesisEnv, is it a blocker for this change?

jveski: lol fixed 😎

internal/controllers/symphony/controller.go
```diff
@@ -179,8 +179,30 @@ func (c *symphonyController) reconcileReverse(ctx context.Context, symph *apiv1.
 func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.Symphony, comps *apiv1.CompositionList) (modified bool, err error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
+	// Prune any annotations that are currently set to empty strings.
+	// This happens before populating other values to enable workflows where an annotation can only be set on (at most) one composition.
```
AYM1607, line 183: Reads a bit obscure, you think it's too much to call out the specifics here?

AYM1607 (approved):

internal/controllers/symphony/controller.go
AYM1607: Works for me :)

PR #479, opened 2025-08-12
Improve overrides docs

jveski:


AYM1607 (approved):

docs/reconciliation/overrides.md
```diff
@@ -33,8 +40,28 @@ annotations:
...
+
+- `eno.azure.io/disable-updates`
+- `eno.azure.io/replace`
+- `eno.azure.io/reconcile-interval`
+
+## Field Manager
+
+Use `pathManagedByEno` to check if Eno currently manages the field specified by `path`. This prevents conflicts with other controllers by conditionally unsetting fields.
+
+This example sets `data.foo` to null only when the field exists but isn't managed by Eno:
```
rayaisaiah, line 64: Maybe explain what setting the field to null does?

docs/reconciliation/overrides.md
```diff
@@ -44,14 +71,18 @@ annotations:
     ]
 ```
 
-## Kubernetes Resource Quantity Comparisons
+Use the same path structure as Kubernetes `metadata.managedFields`. Arrays are typically indexed by key, not numeric index:
+
+- ✅ `self.spec.template.spec.containers[name='myContainer'].image`
+- ❌ `self.spec.template.spec.containers[0].image`
```
rayaisaiah, line 77: Add a note that this works opposite with the `condition` syntax i.e. needs to be indexed like `self.spec.template.spec.containers[0].image`

jveski: Great catch - added a note.

PR #482, opened 2025-08-13
Support scheme in synth sdk

jveski:
Pass a scheme instead of manually setting kind/apiVersion on every object!


AYM1607 (approved):
There's flakes, but they seem unrelated.

PR #496, opened 2025-08-26
Implement resource selector

jveski:
eno-reconciler can now match on specific resources to be reconciled based on their labels. The idea is to allow some reconcilers to watch an overlapping subset of compositions while avoiding conflicts by partitioning individual resources.


jveski:
Not sure why Tingting isn't an approver but she should be - merging anyway.

PR #498, opened 2025-08-28
Fix test flake

jveski:
I don't know why this flake only showed up now, but I can't see any way it could be caused by the PRs that have landed since it started so should be safe to paper over it.


paulgmiller (approved):
handle them race conditions!

PR #499, opened 2025-08-28
Better executor error handling

jveski:
The reason why a synthesizer couldn't be executed by Eno's `executor` process is currently logged but not written anywhere. For the sake of debugging it would be nice to avoid getting the synth pod logs so let's write it to the composition status.


internal/execution/executor.go
```diff
@@ -58,31 +58,40 @@ func (e *Executor) Synthesize(ctx context.Context, env *Env) error {
...
+	if handlerErr != nil {
+		logger.Error(handlerErr, "unable to execute synthesizer")
 
-	var sliceRefs []*apiv1.ResourceSliceRef
-	if resultErr == nil {
-		sliceRefs, err = e.writeSlices(ctx, comp, output)
+		output = &krmv1.ResourceList{Results: []*krmv1.Result{{
+			Message:  "Synthesizer error: " + handlerErr.Error(),
+			Severity: krmv1.ResultSeverityError,
+		}}}
```
paulgmiller, line 72: calling udate compiositions  wth explcity niul slice refs and errOutput then returning eary a little more clear than sharing the updateComposition with success case? 

Kind of aesthetic quibble  but most other errors  are return right away

internal/execution/executor.go
```diff
@@ -58,31 +58,40 @@ func (e *Executor) Synthesize(ctx context.Context, env *Env) error {
...
+	// The handler error is either the error result emitted by the synthesizer
+	// or the handler's underlying "transport" error e.g. exec, json, etc.
+	if handlerErr != nil {
+		logger.Error(handlerErr, "unable to execute synthesizer")
 
-	var sliceRefs []*apiv1.ResourceSliceRef
-	if resultErr == nil {
-		sliceRefs, err = e.writeSlices(ctx, comp, output)
+		output = &krmv1.ResourceList{Results: []*krmv1.Result{{
+			Message:  "Synthesizer error: " + handlerErr.Error(),
```
paulgmiller, line 70: should we truncate here or rely on handler? Can we blow up a custom resoruce with a huge error?

internal/execution/executor.go
```diff
@@ -58,31 +58,40 @@ func (e *Executor) Synthesize(ctx context.Context, env *Env) error {
...
+			Severity: krmv1.ResultSeverityError,
+		}}}
+	} else {
+		handlerErr = findResultError(output)
```
paulgmiller, line 74: why still preflight if we got an error here? Leave a comment if theres a reason why? 

Could also make this have its own update composition and resturn early but it was like this before.

internal/execution/handler.go
```diff
@@ -50,14 +53,19 @@ func NewExecHandler() SynthesizerHandle {
 		cmd.Stderr = os.Stdout // logger uses stderr, so use stdout to avoid race condition
 		cmd.Stdout = stdout
 		err = cmd.Run()
+		if errors.Is(err, exec.ErrNotFound) {
+			return nil, err
+		}
 		if err != nil {
-			return nil, fmt.Errorf("executing command: %w", err)
+			logger.V(0).Info("stdout buffer contents", "stdout", stdout)
```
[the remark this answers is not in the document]
jveski: the logger will truncate or at least the forwarding infra

internal/execution/executor.go
jveski: IMO it's fine to trust handler here since it's an implementation detail of the same package, just abstracted out for testing. Worst case the synthesizer update fails because it's too large which is fine

PR #502, opened 2025-09-02
Add composition status logger

jveski:
Moves the current status transition logs to a separate controller loop that is capable of periodically logging composition status when nothing has changed.

The idea is to keep a fresh representation of each composition's status in indexed log storage.


jveski:
Accidentally deleted the branch lol

Issue #350, opened 2025-04-24
Support per-resource label selector in eno-reconciler process


jveski:
It should be possible for two distinct eno-reconciler processes to reconcile a subset of a single composition's resources using label selectors.

This will be tricky to implement if we want to support moving resources smoothly between reconcilers.

PR #509, opened 2025-09-10
Use resource filter expression instead of a label selector.

AYM1607:
This allows matching on both resource and composition labels. Also move the filter check to the cache so we can avoid keeping unwanted resources in memory.


AYM1607:
Closes #350

jveski (approved):

PR #511, opened 2025-09-10
Support ordered deletion

jveski:
WIP

### DIFF
--- a/docs/reconciliation.md
+++ b/docs/reconciliation.md
@@ -35,17 +35,28 @@ metadata:
     eno.azure.io/disable-reconciliation: "true"
 ```
 
-### Automatic Deletion
+### Deletion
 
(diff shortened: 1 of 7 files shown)


docs/reconciliation.md
```diff
@@ -122,6 +133,18 @@ metadata:
...
+
+By default, resources are deleted without regard to their readiness groups. To enable ordered deletion, use this annotation:
+
+```yaml
+metadata:
+  annotations:
+    eno.azure.io/ordered-deletion: "true"
+```
+
+With ordered deletion enabled, a resource's deletion is blocked until all resources in **higher** numbered groups have been deleted (the inverse order of reconciliation).
```
AYM1607, line 146: Do you think there's any scenarios where we'd want something other than reverse readiness-group?

PR #515, opened 2025-09-16
Support embedded struct fields in input structs.

AYM1607:
Makes it easier to share logic between synthesizers that have common inputs.


jveski (approved):
🚀

PR #518, opened 2025-09-17
Support foreground deletion strategy

jveski:
Setting `eno.azure.io/deletion-strategy: Foreground` will cause eno-reconciler to wait until the resource is done deleting before marking its status as `Deleted==true`.

This is useful for blocking composition deletion until some resources are completely gone.

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -402,7 +402,7 @@ func (c *Controller) requeue(logger logr.Logger, resource *resource.Snapshot, re
 		return ctrl.Result{RequeueAfter: wait.Jitter(c.readinessPollInterval, 0.1)}, nil
 	}
 
-	if resource == nil || (resource.Deleted() && resource.Disable) || resource.ReconcileInterval == nil {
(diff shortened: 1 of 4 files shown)


internal/controllers/reconciliation/controller.go
```diff
@@ -402,7 +402,7 @@ func (c *Controller) requeue(logger logr.Logger, resource *resource.Snapshot, re
 		return ctrl.Result{RequeueAfter: wait.Jitter(c.readinessPollInterval, 0.1)}, nil
 	}
 
-	if resource == nil || (resource.Deleted() && !resource.Disable) || resource.ReconcileInterval == nil {
+	if resource == nil || resource.ReconcileInterval == nil {
```
AYM1607, line 405: Why is the second term no longer required?

jveski: It turns out this change didn't completely make sense. It fixed the foreground deletion test by requeueing  in cases where arguably we shouldn't.

I pushed a commit to keep the current logic, but add a special case: use the readiness polling interval when requeueing during foreground deletion.

AYM1607 (approved):

PR #520, opened 2025-09-22
Fix: Update AnnotateOverrides to Append New Overrides on Existing Overrides

rayaisaiah:
Fixes a bug in the the `AnnotateOverrides` function where if the "eno.azure.io/overrides" annotation existed, the function call would append a duplicate override to the list of overrides instead of appending new overrides to existing overrides.


paulgmiller (approved):
Thanks for the tests

jveski (approved):

PR #526, opened 2025-10-08
Support eno.azure.io/ignore-side-effects on input resources

jveski:
Inputs can now include the ignore-side-effects annotation to override the setting inherited from the composition.

- If the composition's value is true and input's is false, changing this particular input will cause synthesis
- If the composition's value is false (unset) and the input's is true, changes to this particular input will __not__ cause synthesis


ruinan-liu (commented):
I've rebased and gave this binary bits to partner team for testing.

api/v1/composition.go
```diff
@@ -210,6 +214,9 @@ func (i *InputRevisions) Less(b InputRevisions) bool {
 		return *i.Revision < *b.Revision
 	}
 	if i.ResourceVersion == b.ResourceVersion {
+		if i.IgnoreSideEffects != nil && b.IgnoreSideEffects != nil && *i.IgnoreSideEffects != *b.IgnoreSideEffects {
+			return *i.IgnoreSideEffects && !*b.IgnoreSideEffects
+		}
```
xiazhan: Why do we need this?

xiazhan: nvm

xiazhan (approved):

api/v1/composition.go
```diff
@@ -210,6 +214,9 @@ func (i *InputRevisions) Less(b InputRevisions) bool {
 		return *i.Revision < *b.Revision
 	}
 	if i.ResourceVersion == b.ResourceVersion {
+		if i.IgnoreSideEffects != nil && b.IgnoreSideEffects != nil && *i.IgnoreSideEffects != *b.IgnoreSideEffects {
```
xiazhan: nit: this condition can be spared.

xiazhan: Please also add comments for this piece of code.

ruinan-liu: Done, removed the check and added comments for explaination.

ruinan-liu (commented):
Replied to comments

api/v1/composition.go
```diff
@@ -210,6 +220,9 @@ func (i *InputRevisions) Less(b InputRevisions) bool {
 		return *i.Revision < *b.Revision
 	}
 	if i.ResourceVersion == b.ResourceVersion {
+		if i.IgnoreSideEffects != nil && b.IgnoreSideEffects != nil {
+			return *i.IgnoreSideEffects && !*b.IgnoreSideEffects
+		}
```
xiazhan: When IgnoreSideEffect is nil, I assume it's treated as false. Should we replace the whole if sentence with something like:
`return i.IngoreSideEffect != nil && *i.IgnoreSideEffect && (b.IgnoreSideEffect == nil || !*b.IgnoreSideEffect)`

ruinan-liu: Thanks added comment and updated the code.

ruinan-liu (commented):
Responded to comments

internal/controllers/scheduling/op.go
```diff
@@ -272,12 +274,33 @@ func inputChangeCount(synth *apiv1.Synthesizer, a, b []apiv1.InputRevisions) (no
...
+					deferred++
+				} else {
+					nonDeferred++
+				}
+			}
 			continue
 		}
 
 		if br.Less(ar) {
+			if ar.IgnoreSideEffects != nil {
```
xiazhan: When br is less than ar, this condition is always true: ar.IgnoreSideEffects == nil || !*ar.IgnoreSideEffects.

xiazhan: Would it be like?
```
if br.Less(ar) {
        forced++;
}
```

internal/controllers/scheduling/op.go
```diff
@@ -255,7 +257,7 @@ func (r opReason) String() string {
 	}
 }
 
-func inputChangeCount(synth *apiv1.Synthesizer, a, b []apiv1.InputRevisions) (nonDeferred, deferred int) {
+func inputChangeCount(synth *apiv1.Synthesizer, a, b []apiv1.InputRevisions) (nonDeferred, deferred, forced int) {
```
xiazhan: Could we rename them to new/old for better readability?

internal/controllers/scheduling/op.go
ruinan-liu: @xiazhan The `prev.Less(cur)` returning `true` does not gaurantee `cur.IgnoreSideEffects` is nil or false. This condition only holds true in the case of same resource version path through `less()`. When ResourceVersion differ ( this is the normal case when an actual input changes), `Less()` compares integer ResouceVersion and doesn't inspect `IgnoreSideEffects` at all. So `cur.IgnoreSideEffects` can be any of `nil/true/false`.

We need to handle all 3 branches of `cur.IgnoreSideEffects`
* `nil` which is the common case. we count this as a normal deferred/nonDeferred change, but don't increment `forece`. Thie preserves composition-level `ignore-side-effect: true` behavior. `classifyOp` then checks `forced==0` to suppress. 
* true which means the input explicitly opts out of triggering resynthesis and the continue will skip entirely.
* false this menas that the input explicitly forces resynthesis even under composition-level ignore we increment `force++`, then fall through to deferred/nonDeferred.

Simplifying to just `force++` would break 2 things.
1. every normal input change would increment `forced`, effectively disabling the composition-level `ignore-side-effects` feature. and 
2. inputs with `IgnoreSideEffects` would still trigger resynthesis instead of being skipped.

internal/controllers/scheduling/op.go
ruinan-liu: Done, thanks

ruinan-liu (commented):
Respond to comments.

xiazhan (approved):

PR #530, opened 2025-10-17
remove crd dependency on deletion

paulgmiller:
Create/Update CR->CRD dependency makes sense but on deletion the crd just triggers deletion of the CRs which we think is fine and no dependecy is needed from CRD->CR

Retaining the CR->CRD dependency during deletion can create a deadlock if a CR is in a lower deletion group.

### DIFF
--- a/internal/resource/tree.go
+++ b/internal/resource/tree.go
@@ -139,7 +139,7 @@ func (t *tree) Get(key Ref) (res *Resource, visible bool, found bool) {
 	if !ok {
 		return nil, false, false
 	}
-	//debug logging on what we're blocked on might help future issues.
+	//TODO: debug logging on what we're blocked on might help future issues.
 	return idx.Resource, (!idx.Backtracks() && len(idx.PendingDependencies) == 0), true
 }
 


internal/resource/tree.go
```diff
@@ -137,6 +143,7 @@ func (t *tree) Get(key Ref) (res *Resource, visible bool, found bool) {
 	if !ok {
 		return nil, false, false
 	}
+	//debug logging on what we're blocked on might help future issues.
```
jveski: we definitely need something to help find deadlocks. All of the obvious places to add logging would be extremely noisy - would it have been useful in your case to expose the `tree` as json on the loopback interface or something?

I'm thinking a debugging workflow like port-forward to eno-controller and `curl localhost:8000/tree/<compositionname>`

internal/resource/tree.go
```diff
@@ -92,11 +92,17 @@ func (b *treeBuilder) Build() *tree {
...
 		t.byManiRef[idx.Resource.ManifestRef] = idx
 
-		// CRs are dependent on their CRDs
+		// CRs are dependent on their CRDs, except in delete where its reversed
+		// can this introduce circular dependencies if readines/deltion groups say opposite?
 		crd, ok := b.byDefiningGK[idx.Resource.GVK.GroupKind()]
 		if ok {
-			idx.PendingDependencies[crd.Resource.Ref] = struct{}{}
-			crd.Dependents[idx.Resource.Ref] = idx
+			if idx.Resource.compositionDeleted {
```
jveski: any reason to invert cr/crd ordering rather than ignoring it during deletion?

paulgmiller: I assume trying to delete a CRD while CR's still exist will just fail so why not wait? You could say the same thing about creation. Eventually the crd will be applied and then the CR will go so either we should ignore for both or obey ordering for both?

paulgmiller: Comment updated and UT added. Feel free to make suggestions to comment.

jveski (approved):

PR #538, opened 2025-11-07
[eno] Added telemetry library and logging for synthesizer in eno

ruinan-liu:
Added below fields to EnoEvents table

synthName, synthUrl, synthCreationTimestamp

### DIFF
--- a/cmd/eno-controller/main.go
+++ b/cmd/eno-controller/main.go
@@ -70,8 +70,8 @@ func runController() error {
 			Rest: ctrl.GetConfigOrDie(),
 		}
 	)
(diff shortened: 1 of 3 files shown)


internal/logging/composition.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return fields
+}
+
+// compositionEventType determines the event type for composition logging
+func compositionEventType(comp *apiv1.Composition) string {
+	if comp.Status.Simplified == nil {
+		return "status_update"
+	}
+	return "status_update"
```
xiazhan: It only has "status_update". How about deleting and creating?

ruinan-liu: Added Status_created, status_deleting, and status_deleted

xiazhan (approved):

PR #542, opened 2025-12-03
migrate ssa ownership

ruinan-liu:


internal/controllers/reconciliation/controller.go
```diff
@@ -472,6 +492,93 @@ func summarizeError(err error) string {
 	}
 }
 
+// MigrateOwnership ensures eno owns all fields under the specified scope by performing
+// a Server side apply with force. Returns true if migration was performed, false if already owned
+func (c *Controller) MigrateOwnership(ctx context.Context, resource *unstructured.Unstructured, scope string) (bool, error) {
```
jveski: This logic seems essentially correct, but there's a lot of duplicate code. Most of it is already defined in the resource package for the other managedFields interactions.

We should be able to split the existing managedFields metadata into two parts: one to run before the dryrun (with all of the logic that doesn't require the dryrun result), and one to run after with whatever logic is remaining. No need to reimplement in the reconciliation pkg.

PR #548, opened 2025-12-10
Add error logging and progress tracking to synthesis executor

tilnl:
Add error logging and progress tracking to executor. 

image

### DIFF
--- a/internal/execution/executor.go
+++ b/internal/execution/executor.go
@@ -28,7 +28,11 @@ type Executor struct {
 }
 
 func (e *Executor) Synthesize(ctx context.Context, env *Env) error {
-	logger := logr.FromContextOrDiscard(ctx).WithValues("synthesisUUID", env.SynthesisUUID)
+	logger := logr.FromContextOrDiscard(ctx).WithValues(
+		"synthesisUUID", env.SynthesisUUID,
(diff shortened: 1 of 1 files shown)


internal/execution/executor.go
```diff
@@ -29,35 +29,42 @@ type Executor struct {
...
 func (e *Executor) Synthesize(ctx context.Context, env *Env) error {
 	logger := logr.FromContextOrDiscard(ctx)
+	logger.Info("starting synthesis", "synthesisuuid", env.SynthesisUUID)
 
 	comp := &apiv1.Composition{}
 	comp.Name = env.CompositionName
 	comp.Namespace = env.CompositionNamespace
 	err := e.Reader.Get(ctx, client.ObjectKeyFromObject(comp), comp)
 	if err != nil {
+		logger.Error(err, "unable to fetch composition")
```
ruinan-liu: Is it possible to format the error logs to container which composition and which synthesizer?

ruinan-liu: something like the below?
		logger.Error(err, fmt.Sprintf("failed to get composition. compition Name[%s] Namespace[%s]", comp.GetName(), comp.GetNamespace()))

PR #549, opened 2025-12-11
[eno] Add OperationID and OpeartionOrigin to eno telemetry

ruinan-liu:
Added operationID and operationOrigin to eno-controller and eno-reconciliation operations


internal/logging/composition.go
```diff
@@ -22,7 +22,6 @@ func NewCompositionStatusLogger(mgr ctrl.Manager, freq time.Duration) error {
 			PredicateFn:     compositionPredicate,
 			ExtractFieldsFn: extractCompositionFields,
 			EventTypeFn:     compositionEventType,
-			MessageFn:       func() string { return "current composition status" },
```
xiazhan: Why is the MessageFn removed here?

PR #583, opened 2026-03-30
[eno] Add default infrastructure resource ordering

ruinan-liu:
Follow helm release, we need to add infrastructure ordering: https://github.com/helm/helm/blob/main/pkg/release/v1/util/kind_sorter.go


internal/resource/resource.go
```diff
@@ -198,6 +198,10 @@ func newResource(ctx context.Context, parsed *unstructured.Unstructured, strict
 		if err != nil {
 			logger.Info("invalid readiness group - ignoring")
 		} else {
+			if rg >= -100 && rg <= -81 {
+				logger.Info("WARNING: user-supplied readiness-group is in Eno reserved range [-100, -81]",
```
xiazhan: Could we make the reserved range larger? We are using most of the reserved range in this pr.

internal/resource/resource.go
```diff
@@ -198,6 +198,10 @@ func newResource(ctx context.Context, parsed *unstructured.Unstructured, strict
 		if err != nil {
 			logger.Info("invalid readiness group - ignoring")
 		} else {
+			if rg >= -100 && rg <= -81 {
```
xiazhan: better to make -100 and -81 constants.

internal/resource/resource.go
```diff
@@ -213,6 +217,10 @@ func newResource(ctx context.Context, parsed *unstructured.Unstructured, strict
 		if err != nil {
 			logger.Info("invalid deletion group - ignoring")
 		} else {
+			if rg >= 81 && rg <= 100 {
+				logger.Info("WARNING: user-supplied deletion-group is in Eno reserved range [81, 100]",
```
xiazhan: Are readiness group all negative and deletion group are positive? We probably need to define clearer relationship among readiness group and deleting group.  I feel deletion group is kind of redundant.

internal/resource/kind_ordering.go
```diff
@@ -0,0 +1,38 @@
...
+	"PodSecurityPolicy":        -98,
+	"PodDisruptionBudget":      -98,
+	"ServiceAccount":           -97,
+	"Secret":                   -96,
+	"SecretList":               -96,
+	"ConfigMap":                -96,
+	"StorageClass":             -95,
+	"PersistentVolume":         -94,
+	"PersistentVolumeClaim":    -93,
+	"CustomResourceDefinition": -92,
```
tilnl, line 28: CRD is before controller deployment by default right?

internal/resource/kind_ordering.go
```diff
@@ -0,0 +1,38 @@
...
+// readiness groups. We intentionally key on Kind alone: any resource
+// whose Kind matches one of these names is treated as infrastructure
+// and reconciled first, regardless of its API group.
+//
+// User-supplied readiness/deletion groups must be >= -80.
+// Values in [-100, -81] are reserved for Eno-managed infrastructure defaults.
+// Deletion groups are the negation of the create groups.
+// Order derived from Helm's InstallOrder/UninstallOrder:
+// https://github.com/helm/helm/blob/main/pkg/release/v1/util/kind_sorter.go
+var managedCreateOrder = map[string]int{
```
tilnl, line 13: The defaults assume infra kinds are always independent and should go first. 
But in multi-component compositions, we may want some deployment ready first before other infra kinds (CRDs/secrets). 
e.g. Get etcd/apiserver deployment ready before fleet CRD. How should users handle this? Should users apply `readiness-group` annotations to all components explicitly? Or should users use lower values (e.g. `-100`) on Deployments to push them before the CRD default of `-92` — even though that puts user values in the reserved range?

internal/resource/resource.go
ruinan-liu: Done, made the edit to make the range [-100, -60]

internal/resource/resource.go
ruinan-liu: Updated, thanks

internal/resource/resource.go
ruinan-liu: on 1) readiness groups aren't required to be negative and deletion group aren't required to be positive. Any integer is allowed. The only place these matters is Eno's reserved range for managed infra kinds, where deletion default is just the negation readiness default. User supplied values can be anything outside that reserved range. 

on 2) combining the readiness groups and deletion groups to one I think we need to be more careful. The two annotations control two distinct phase (creation readiness vs teardown ordering), and keep them separate lets users express orderings that aren't strict inverse of each other. Helm also maintain a separate InstallOrder and UninstallOrder for the same reason. I will make sure to also update the doc to call this out as well.

internal/resource/kind_ordering.go
ruinan-liu: Yes, I did not add deployment to the managed kinds because we could have dep1 be ready before dep2 so user needs to have this flexibility. The default deployment will be 0 which is higher then the current CRD.

internal/resource/kind_ordering.go
ruinan-liu: Thanks Tingting for raising this concern. I think the current behavior on main (without this PR) is that both CRDs and Deployments get `readinessgroup = 0` by default. There is no managed-kind defaulting today. So the etcd-apiserver-before-crd scenario already requires the user to explicitly annotate (e.g. Deployment -10, CRD 0) to get any non-trivial ordering. 

For the etcd/apiserver-before-fleet-CRD scneario, this PR does introduce one extra annotation compared to today. On main, you'd onlly need `readiness-group:-10` on the deployment (CRD defaults to 0). With this PR, you'd also need `readiness-group:0` on the fleet CRD to lift it above the `Deployment's -10`, since CRD now defaults to `-92`.

ruinan-liu (commented):
Reply to comments

xiazhan (approved):

PR #585, opened 2026-04-06
[eno][part2] Adding composition creation order for creation/update

ruinan-liu:
Adding controller logic for using DependsOn on creation flow.


internal/controllers/symphony/controller.go
```diff
@@ -225,8 +225,22 @@ func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.
...
+	// NEW: Build map of synthesizer name -> existing composition for dependency resolution
+	compBySynth := make(map[string]*apiv1.Composition, len(comps.Items))
+	for i := range comps.Items {
+		compBySynth[comps.Items[i].Spec.Synthesizer.Name] = &comps.Items[i]
+	}
+
+	// Sort variations so dependencies come first; detect cycles as a byproduct
+	sortedVariations, cyclicSynths := topoSortVariations(symph.Spec.Variations)
+	for synthName := range cyclicSynths {
+		logger.Info("circular dependency detected in variation, skipping",
```
xiazhan: It seems that we didn't skip. I wonder whether we should fail the whole operation.

internal/controllers/symphony/controller.go
```diff
@@ -225,8 +225,22 @@ func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.
...
-	// Sync forward (create/update)
-	for _, variation := range symph.Spec.Variations {
+	// NEW: Build map of synthesizer name -> existing composition for dependency resolution
+	compBySynth := make(map[string]*apiv1.Composition, len(comps.Items))
+	for i := range comps.Items {
+		compBySynth[comps.Items[i].Spec.Synthesizer.Name] = &comps.Items[i]
+	}
+
+	// Sort variations so dependencies come first; detect cycles as a byproduct
+	sortedVariations, cyclicSynths := topoSortVariations(symph.Spec.Variations)
```
xiazhan, line 235: Thanks for adding it in this change.

internal/controllers/symphony/controller.go
```diff
@@ -235,6 +249,37 @@ func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.
...
+		// Resolve variation dependencies to composition dependencies. If the dependent composition does not exist yet
+		// DO NOT CREATE the current composition as it might lead to race condition and ordering not being respected
+
+		deps, allresolved := resolveVariationDeps(variation.DependsOn, compBySynth)
+		for _, dep := range variation.DependsOn {
+			if dep.Synthesizer == "" {
+				logger.Error(fmt.Errorf("No Variation Dependency Synthesizer"),
+					"Error: variation dependency has no synthesizer set, dependency will be ignored",
+					"synthesizerName", variation.Synthesizer.Name)
+			}
```
xiazhan, line 263: Should we fail the operation? So we don't need to worry about how to ignore it?

internal/controllers/composition/controller.go
```diff
@@ -323,6 +323,11 @@ func buildSimplifiedStatus(synth *apiv1.Synthesizer, comp *apiv1.Composition) *a
 	}
 
 	if comp.Status.CurrentSynthesis == nil && comp.Status.InFlightSynthesis == nil {
+		// This means that dependencies exists and no synthesis started showing WaitingOnDependencies
```
xiazhan, line 326: We removed a lot of changes in this controller compared to the whole pr? Is it expected?

internal/controllers/scheduling/controller.go
```diff
@@ -114,9 +115,48 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	// Reset the compositionHealth metric before iterating through compositions
 	compositionHealth.Reset()
 
+	// Build readiness index and composition map for dependency checking
+	readySet := buildReadySet(comps)
+	sortedComps, cyclicSet := topoSortCompositions(comps.Items)
```
xiazhan, line 122: If there is a circle, I feel we shouldn't proceed.

internal/controllers/symphony/controller.go
ruinan-liu: Hey Allan, currently if there is a cycle, we will handle that in the composition's status here

	// Pre-loop: set/clear circular dependency status on all compositions.
	// Cyclic compositions are excluded from sortedComps by Kahn's algorithm,
	// so we must handle them here before the main loop.
	for _, comp := range comps.Items {
		comp := comp
		if comp.DeletionTimestamp != nil || len(comp.Spec.DependsOn) == 0 {
			continue
		}
...
From my understanding, we currently don't have the methods for propagating information error information from composition to symphony and currently we are handling this composition. The operation will fail and see that  the reason is circular dependency. 

The reason we don't skip is becaue the `sortedVariations` won't contain the cyclic variations due to topological sort. So we will not process them and the all reconciled will not ready. 
But I think we should introduce more robust error propagation here.

internal/controllers/symphony/controller.go
ruinan-liu: Thanks for pointing this out, after looking at this piece of code more carefully, it actually has a subtle issue which is if a `variation`'s dependency does not have a synthesizer, then the `deps, allresolved := resolveVariationDeps(validDeps, compBySynth)` will always be `false` and stuck the whole operation because later in `resolveVariationDeps` the line `target, ok := compBySynth[""]` would return `ok=false` so `allresolved` will always be false and we will never create the compositions. I changed our current code so that we check on properly created synthesizers as a best effort basis. Currently `Symphony` does not have a good way to propogate error/information up given the only condition is `reconciled`.

Regarding failing the operation: SymphonyStatus only exposes timestamps (Synthesized/Reconciled/Ready), not error conditions, so there's currently no way to surface validation errors to consumers. As a future improvement, we could add a validating webhook to the eno-controller to reject symphonies where dependencies reference empty or non-existent synthesizers at admission time.

internal/controllers/composition/controller.go
ruinan-liu: Hey Allan, yes. This is part2 which we mainly focus on the creation/update order. For deletion and e2e tests I will have a part 3.

internal/controllers/scheduling/controller.go
ruinan-liu: Hey Allan, since compositions are always created through a symphony and cross-symphony dependenceis aren't a thing, the symphony controller's topo sort already prevents cyclic compositiosn from being created. The cycle detec in the scheduling controller is a defensive fallback that should not trigger under normal operation. Ideally this validation would move to a webhook to reject the invalid spec at admission time. For now, the scheduling controller check serves as a safety net in  case the symphony controller's cycle detection is bypassed. We can remove this in the future when we having better ways to report errors.

ruinan-liu (commented):
replied to comments

internal/controllers/scheduling/controller.go
tilnl: Although AKS cluster always define a symphony, composition could be created directly without symphony.

internal/controllers/scheduling/controller.go
```diff
@@ -114,9 +115,48 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// Cyclic compositions are excluded from sortedComps by Kahn's algorithm,
+	// so we must handle them here before the main loop.
+	for _, comp := range comps.Items {
+		comp := comp
+		if comp.DeletionTimestamp != nil || len(comp.Spec.DependsOn) == 0 {
+			continue
+		}
+		compKey := path.Join(comp.GetNamespace(), comp.GetName())
+		if cyclicSet[compKey] {
+			logger.Info("circular dependency detected, skipping composition synthesis",
```
tilnl: should the log be error or warning?

internal/controllers/symphony/controller.go
tilnl: should the log be error or warning

internal/toposort/toposort_test.go
```diff
@@ -0,0 +1,149 @@
...
+			expectedOrder:  []string{},
+			expectedCyclic: []string{"a", "b", "c"},
+		},
+		{
+			name: "dep references non-existent node",
+			items: []node{
+				{name: "a", deps: []string{"phantom"}},
+			},
+			expectedOrder:  []string{},
+			expectedCyclic: []string{"a"},
```
tilnl: Is a non-existent dependency treated as a cyclic dependency?

api/v1/composition.go
```diff
@@ -9,13 +9,15 @@ import (
...
-	WaitingOnDependentsReason   string = "WaitingOnDependents"
-	WaitingOnDependenciesReason string = "WaitingOnDependencies"
+	enoAzureOperationIDKey                   = "eno.azure.io/operationID"
+	enoAzureOperationOrigin                  = "eno.azure.io/operationOrigin"
+	OperationIdKey                    string = "operationID"
+	OperationOrigionKey               string = "operationOrigin"
+	CircularDependencyReason          string = "CircularDependency"
+	WaitingOnDependentsReason         string = "WaitingOnDependents"
+	WaitingOnDependenciesReason       string = "WaitingOnDependencies"
+	WaitingOnDependencyNotFoundReason string = "DependencyNotFound"
```
tilnl, line 19: Is DependencyNotFound used?

internal/controllers/symphony/controller.go
```diff
@@ -235,6 +249,40 @@ func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.
...
+		idx := slices.IndexFunc(comps.Items, func(existing apiv1.Composition) bool {
+			return existing.Spec.Synthesizer.Name == variation.Synthesizer.Name
+		})
+
+		// Safety fallback: with topological sort this should not trigger since
+		// dependencies are processed first, but guards against edge cases like
+		// a failed Create earlier in the loop.
+		if !allresolved {
+			if idx == -1 {
+				logger.Info("skipping composition creation: not all synthesizer-based dependencies resolved yet",
```
xiazhan: nit: This is an error.

internal/controllers/symphony/controller.go
```diff
@@ -235,6 +249,40 @@ func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.
...
+					"synthesizerName", variation.Synthesizer.Name)
+			} else {
+				logger.Info("skipping composition update: not all synthesizer-based dependencies resolved yet",
```
xiazhan: should an error.

xiazhan:
### Code review

Found 3 issues:

1. **Compositions referencing non-existent dependencies are silently treated as "not ready" rather than flagged as "not found".** The `WaitingOnDependencyNotFoundReason` constant is defined but never used. In `scheduling/controller.go`, when a dependency key is not in `readySet`, the code always sets `DependencyNotReady` — but the dependency composition might not exist at all. There is no distinction between "exists but not ready" and "does not exist". This makes debugging harder and the `DependencyNotFound` reason dead code.

https://github.com/Azure/eno/blob/4a465d4c4a9117520e145caccee977c45aa2f03c/internal/controllers/scheduling/controller.go#L174-L191

2. **The `WaitingOnDependencies` status update comparison only checks `Reason`, not `BlockedBy` contents.** If the set of blocking dependencies changes (e.g., one dependency becomes ready but another is still blocked), the `BlockedBy` list in the status will be stale because the condition `comp.Status.DependencyStatus.Reason != newStatus.Reason` is already false (both are `WaitingOnDependencies`), so the patch is skipped. The status will only update once all deps are ready or the reason changes entirely.

https://github.com/Azure/eno/blob/4a465d4c4a9117520e145caccee977c45aa2f03c/internal/controllers/scheduling/controller.go#L195-L202

3. **`resolveVariationDeps` treats an empty `Synthesizer` field as "unresolved" (returns `allResolved=false`), but the caller already filters these out.** This creates a subtle inconsistency: the caller in `symphony/controller.go` strips empty-synthesizer deps into `validDeps` before calling `resolveVariationDeps`, but `resolveVariationDeps` itself also handles empty synthesizer by failing lookup (empty string won't be in `compBySynth`). The double handling is benign but the function's contract is unclear — if it's supposed to receive pre-validated deps, the empty-string guard in `topoSortVariations` and the caller's filter are redundant. More importantly, the test `"dep with empty synthesizer is unresolved"` tests `resolveVariationDeps` directly with an empty synthesizer, which the real call path would never hit.

https://github.com/Azure/eno/blob/4a465d4c4a9117520e145caccee977c45aa2f03c/internal/controllers/symphony/controller.go#L257-L270

xiazhan:
**Addendum to code review — one more issue:**

4. **Non-existent dependencies are misclassified as circular dependencies.** In `TopologySort`, if a composition references a dependency that does not exist in the item set (e.g., typo in `DependsOn`, deleted composition, or not-yet-created), the referencing item's in-degree is incremented but the missing key is never enqueued (no entry in `byKey`). The item's in-degree never reaches 0, so it lands in the `cyclic` set — even though there is no actual cycle. The scheduling controller then labels it with `CircularDependencyReason`, which is incorrect and misleading. The user sees "circular dependency" when the real problem is "dependency not found."

https://github.com/Azure/eno/blob/4a465d4c4a9117520e145caccee977c45aa2f03c/internal/toposort/toposort.go#L20-L25

**Fix:** Before incrementing in-degree for a dep key, check if it exists in the item set. If not, track it separately as "unresolved" rather than letting it fall into the cyclic bucket.

internal/controllers/scheduling/controller.go
ruinan-liu: We will add status section in Symphony CR to track the status of the symphony for quick failure

internal/controllers/scheduling/controller.go
ruinan-liu: Thanks changed this to be error

internal/toposort/toposort_test.go
ruinan-liu: Fixed, thanks

api/v1/composition.go
ruinan-liu: Fixed, thanks

ruinan-liu:
Respond to comments:
1. Compositions referencing non-existent dependencies are silently treated as "not ready" rather than flagged as "not found".
**This is fixed in the code**

3.  The WaitingOnDependencies status update comparison only checks Reason, not BlockedBy contents.
This is fixed by using
```
if comp.Status.DependencyStatus != nil && equality.Semantic.DeepEqual(*comp.Status.DependencyStatus, *newStatus) {
		return false, nil // modified=false, err = nil
	}
```
4. resolveVariationDeps treats an empty Synthesizer field as "unresolved" (returns allResolved=false), but the caller already filters these out.
`resolveVariationDeps` implicitly rejects empty-synthesier dependencies( they won't match any key in `compBySynth`), but the caller in `reconcileForward` already explicitly filters them out into `validDeps` before calling it. This double handling is benign - both layers produce the correct result, and the redundancy provides defense in depth if a future call site skips pre-filtering. The tests `dep with empty synthesizer is unresolved` exercise an unreachable path forn the current caller, but it still documents the function's standalone contract. Removing either guard would tighten coupling for zero behavior benefit, so this is intentionally left as-is. 

5. Non-existent dependencies are misclassified as circular dependencies.
Fixed in the test.

ruinan-liu (commented):
Responded to comments

ruinan-liu (commented):
Thanks, replied to comments

tilnl (approved):

internal/controllers/symphony/controller.go
```diff
@@ -225,8 +225,23 @@ func (c *symphonyController) reconcileForward(ctx context.Context, symph *apiv1.
...
+	// NEW: Build map of synthesizer name -> existing composition for dependency resolution
+	compBySynth := make(map[string]*apiv1.Composition, len(comps.Items))
+	for i := range comps.Items {
+		compBySynth[comps.Items[i].Spec.Synthesizer.Name] = &comps.Items[i]
+	}
+
+	// Sort variations so dependencies come first; detect cycles as a byproduct
+	sortedVariations, cyclicSynths := topoSortVariations(symph.Spec.Variations)
+	for synthName := range cyclicSynths {
+		logger.Error(fmt.Errorf("circular dependency detected in variation"),
```
tilnl, line 237: for new symphony, if there is circular dependency, the composition will be skipped ? Will there be status in symphony

ruinan-liu: Thanks Tingting, yes if it is circular dependency then we will not create the composition. I don't think it will be an issue now because the dependency is straight forward. We currently have an issue planned: https://github.com/Azure/eno/issues/586. 

Currently symphony CR do not have any status or do we expose any status for quick validation errors. We will be working on surfacing those errors in this issue

ruinan-liu (commented):
REplied to comments

PR #590, opened 2026-04-15
[eno][part3] Adding Deletion ordering and unit test

ruinan-liu:
This is part 3 of the PR which I added deletion ordering and e2e


examples/03-helm-shim/Dockerfile
```diff
@@ -1,12 +1,11 @@
 FROM mcr.microsoft.com/devcontainers/go:1.24 AS builder
-WORKDIR /app
 
-ADD go.mod .
-ADD go.sum .
-RUN --mount=type=cache,target=/root/.cache/go-build go mod download
+ENV GOTOOLCHAIN=auto
```
ruinan-liu, line 2: The change in DOCKERFILE is to pass the Smoke test

internal/controllers/composition/controller.go
tilnl: Will it be possible the Symphony is deleted and a dependent composition is stuck then annotation forceDeleteWhenSymphonyGone doesn't work? Since it is checked after hasActiveDependents

internal/controllers/composition/controller.go
```diff
@@ -166,19 +241,12 @@ func (c *compositionController) reconcileDeletedComposition(ctx context.Context,
...
 		if syn.Reconciled == nil {
-			// If this is an addon composition whose owning Symphony is already gone,
-			// force-remove the finalizer so the composition doesn't get stuck forever.
-			if c.shouldForceRemoveFinalizer(ctx, comp) {
-				logger.Info("force removing finalizer for composition because owning symphony is gone and composition is being marked force delete",
-					"compositionName", comp.Name, "compositionNamespace", comp.Namespace)
-			} else {
-				logger.Info("refusing to remove composition finalizer because it is still being reconciled")
-				return ctrl.Result{}, nil
-			}
```
xiazhan, line 177: Are you saying we'll enforce that the overlay apiserver is deleted after the addons are removed?
Is there any chance that the addons cannot be deleted even if the overlay apiserver exists?

xiazhan: Ok. It was moved to the beginning of the function.

internal/controllers/composition/controller.go
```diff
@@ -144,6 +182,43 @@ func (c *compositionController) Reconcile(ctx context.Context, req ctrl.Request)
...
+	}
+
+	// Once the given compositions's dependents are deleted then we can clear the dependency status
+	// Note that once we clear the hasActiveDependents step we will proceed to deletion. This clear step
+	// is just for the correctness  step during this breif window
+	if comp.Status.DependencyStatus != nil {
+		if _, err = c.clearDependencyStatus(ctx, comp); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: wait.Jitter(time.Second*5, 0.1)}, nil // Explicit requeue to continue deletion after clearing dependency status.
```
xiazhan: Why do we need to do it explicitly? clearDependencyStatus will make the composition updated, thus make it reconciled again.

internal/controllers/composition/controller.go
```diff
@@ -41,13 +46,46 @@ func NewController(mgr ctrl.Manager, podTimeout time.Duration) error {
...
 		For(&apiv1.Composition{}).
 		WatchesRawSource(source.Kind(mgr.GetCache(), &apiv1.Synthesizer{}, c.newSynthEventHandler())).
+		WatchesRawSource(source.Kind(mgr.GetCache(), &apiv1.Composition{}, c.newDependencyEventHandler(), depPredicate)).
 		WithLogConstructor(manager.NewLogConstructor(mgr, "compositionController")).
 		Complete(c)
 }
 
+func (c *compositionController) newDependencyEventHandler() handler.TypedEventHandler[*apiv1.Composition, reconcile.Request] {
+	fn := func(ctx context.Context, comp *apiv1.Composition) (reqs []reconcile.Request) {
+		// When a dependency changes (e.g. deleted), notify its dependencies
```
xiazhan: Should we recheck dependency only when a composition's finalizer is removed?

internal/controllers/composition/controller.go
ruinan-liu: Done, thanks Tingting. I moved to the beginning of the deletion block.

internal/controllers/composition/controller.go
ruinan-liu: Yeah, Tingting spotted an issue with this current flow and I moved it to the beginning of the block.

internal/controllers/composition/controller.go
ruinan-liu: Yeah, initially I thought adding that as a safety guard, but I realized that this will trigger 2 events and not one guaranteed. I removed it. Thanks

internal/controllers/composition/controller.go
ruinan-liu: Done, we only check when the finalizer is removed from the object.

ruinan-liu (commented):
reply

xiazhan (approved):

tilnl (approved):

PR #595, opened 2026-04-21
feat(overrides): add CEL valueProgram support for dynamic override values

JinghanFu:
## Summary

Adds a `valueProgram` field to overrides, allowing CEL expressions to dynamically compute override values from the current (live) resource state. Previously, overrides could only set static values. With `valueProgram`, the override value is evaluated at apply time against the actual cluster resource, enabling scenarios like preserving VPA-adjusted resource quantities.

## Motivation

When external actors (e.g., VPA) modify resource fields, Eno's static overrides couldn't reference the current cluster state to decide what value to use. `valueProgram` solves this by letting a CEL expression read from the live resource and return a computed value.

### `internal/resource/mutation/mutation.go`

- Added `ValueProgram cel.Program` field to `Op` struct
- Added `valueProgram` JSON field to `jsonOp` for deserialization
- `UnmarshalJSON`: parses `valueProgram` string into a compiled `cel.Program` via `enocel.Parse()`
- `Apply`: if `ValueProgram` is set, evaluates it against the current resource state to resolve the value dynamically; skips mutation if current is nil or CEL evaluates to null

### `internal/resource/mutation/mutation_test.go`

- Added tests for `valueProgram` evaluation, null handling, and missing current resource scenarios

### `pkg/function/overrides/override.go`

- Added `ValueProgram string` field to `Override` struct
- `validate()`: if `ValueProgram` is non-empty, parses it as CEL to ensure validity

### `pkg/function/overrides/override_test.go`

- Added unit tests for `ValueProgram` validation and `Override.Test()` behavior

### `internal/controllers/reconciliation/overrides_test.go`

- Added integration-level tests exercising `valueProgram` through the full `SnapshotWithOverrides` → `Op.Apply` path

### `docs/overrides.md`

- Added documentation for the `valueProgram` field with usage examples

## Testing

- All existing and new unit tests pass (`go test ./internal/resource/mutation/... ./pkg/function/overrides/... ./internal/cel/...`)

### DIFF
--- a/e2e/default_ordering_test.go
+++ b/e2e/default_ordering_test.go
@@ -1,0 +1,225 @@
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	flow "github.com/Azure/go-workflow"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	apiv1 "github.com/Azure/eno/api/v1"
+	fw "github.com/Azure/eno/e2e/framework"
+)
+
+// TestResourceDependencyOrdering validates that Eno correctly handles the
+// dependency ordering between a Secret and a Deployment that mounts it,
+// WITHOUT explicit eno.azure.io/readiness-group annotations.
+//
+// The synthesizer outputs both a Secret and a Deployment (which mounts
+// the Secret as a volume). The test verifies:
+//   - Both resources are created successfully
+//   - The Deployment becomes available (at least one ready pod)
+//   - Pods have zero restarts (no CrashLoopBackOff)
+//   - No container errors (waiting/terminated with failure)
+func TestResourceDependencyOrdering(t *testing.T) {
+	t.Parallel()
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	cli := fw.NewClient(t)
+
+	synthName := fw.UniqueName("dep-order-synth")
+	compName := fw.UniqueName("dep-order-comp")
+	secretName := fw.UniqueName("dep-order-secret")
+	deployName := fw.UniqueName("dep-order-deploy")
+
+	// Secret with NO readiness-group annotation.
+	secret := &corev1.Secret{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "v1", Kind: "Secret"},
+		ObjectMeta: metav1.ObjectMeta{Name: secretName, Namespace: "default"},
+		StringData: map[string]string{"token": "test-value-123"},
+	}
+
+	// Deployment that mounts the Secret as a volume — NO readiness-group annotation.
+	replicas := int32(1)
+	deploy := &appsv1.Deployment{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "apps/v1", Kind: "Deployment"},
+		ObjectMeta: metav1.ObjectMeta{Name: deployName, Namespace: "default"},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": deployName},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"app": deployName},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Name:    "test",
+						Image:   "docker.io/busybox:1.36.1",
+						Command: []string{"sh", "-c", "cat /etc/secret-vol/token && sleep 3600"},
+						VolumeMounts: []corev1.VolumeMount{{
+							Name:      "secret-vol",
+							MountPath: "/etc/secret-vol",
+							ReadOnly:  true,
+						}},
+					}},
+					Volumes: []corev1.Volume{{
+						Name: "secret-vol",
+						VolumeSource: corev1.VolumeSource{
+							Secret: &corev1.SecretVolumeSource{
+								SecretName: secretName,
+							},
+						},
+					}},
+				},
+			},
+		},
+	}
+
+	// Synthesizer outputs both Secret and Deployment.
+	// Eno should automatically resolve the dependency ordering.
+	synth := fw.NewMinimalSynthesizer(synthName, fw.WithCommand(fw.ToCommand(secret, deploy)))
+	comp := fw.NewComposition(compName, "default",
+		fw.WithSynthesizerRefs(apiv1.SynthesizerRef{Name: synthName}))
+	compKey := types.NamespacedName{Name: compName, Namespace: "default"}
+
+	// --- Workflow steps ---
+
+	createSynthesizer := fw.CreateStep(t, "createSynthesizer", cli, synth)
+
+	createComposition := fw.CreateStep(t, "createComposition", cli, comp)
+
+	waitCompositionReady := flow.Func("waitCompositionReady", func(ctx context.Context) error {
+		fw.WaitForCompositionReady(t, ctx, cli, compKey, 4*time.Minute)
+		t.Log("composition is Ready")
+		return nil
+	})
+
+	verifySecret := flow.Func("verifySecret", func(ctx context.Context) error {
+		s := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{Name: secretName, Namespace: "default"},
+		}
+		fw.WaitForResourceExists(t, ctx, cli, s, 30*time.Second)
+		assert.Equal(t, "test-value-123", string(s.Data["token"]),
+			"secret should contain expected data")
+		t.Logf("secret %s exists with correct data", secretName)
+		return nil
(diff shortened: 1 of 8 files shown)


pkg/function/overrides/override.go
```diff
@@ -16,9 +16,10 @@ import (
 // trying to do type Override = intmut.Op will get you an erro about extending methods.
 // could make it a composition but not going down that path yet
 type Override struct {
-	Path      string `json:"path"`
-	Value     any    `json:"value"`
-	Condition string `json:"condition"`
+	Path         string `json:"path"`
+	Value        any    `json:"value,omitempty"`
```
ruinan-liu: is adding `omitempty` intentinoal?

pkg/function/overrides/override.go
```diff
@@ -55,6 +56,13 @@ func (o *Override) validate() (cel.Program, error) {
 		return nil, fmt.Errorf("failed to create program: %w", err)
 	}
```
ruinan-liu, line 57: nit: If `value` and `valueProgram` are mutually exclusive, can we add a logging here for telemetry?

internal/resource/mutation/mutation.go
```diff
@@ -90,7 +101,26 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
...
+		if err != nil {
+			logger.Error(err, "failed to evaluate value expression", "path", o.Path.String())
+			return StatusInvalidValueProgram, nil
+		}
+		resolvedValue = val.Value()
+		if resolvedValue == nil || resolvedValue == structpb.NullValue_NULL_VALUE {
+			logger.Info("CEL value expression evaluated to null, skipping mutation", "path", o.Path.String())
+			return StatusInactive, nil
+		}
+		logger.Info("resolved CEL value expression", "path", o.Path.String(), "resolvedValue", resolvedValue)
```
ruinan-liu: would logging `resolvedValue` be too sensitive to log in telemetry.

ruinan-liu:
Could we also add a e2e test like in this PR? https://github.com/Azure/eno/pull/590/changes#diff-ff578928676707b352eb1dd2ee6a304cf2887ec8969d83e872ba5b4243b1f930. 
You could ask copilot to generate the tests, it's pretty accurate in writing tests.

ruinan-liu:
For smoke test failures, I have a fix in this PR: https://github.com/Azure/eno/pull/526. I'm currently working on getting it to merge.

ruinan-liu (commented):
Thanks Jinghan for the Pr. I've added some minor comments. Please let me know if you have any question.

pkg/function/overrides/override.go
JinghanFu: Hi, Ruinan. May I know whether you mean checking the cases for mutually exclusive and then return nil, Errorf("error message") here (consistent with all other checks in this function) or printing this out by fmt.Printf("error message")? Thank you!

ruinan-liu: Yeah sorry maybe not here, but it would be better if we can add a logging in telemetry saying that we are currently using valueProgram instead of Value. 

Not a blocker, if it's too complicated we don't need this.

ruinan-liu (commented):
reply to comments

pkg/function/overrides/override.go
JinghanFu: Thank you, Ruinan! I put the logging in telemetry in mutation.go and they look like:
`if o.ValueProgram != nil { 
...
logger.Info("override using valueProgram (resolved CEL value expression)", "path", o.Path.String())
} else {
logger.Info("override using static default value", "path", o.Path.String())
}
`

pkg/function/overrides/override.go
```diff
@@ -16,9 +16,10 @@ import (
 // trying to do type Override = intmut.Op will get you an erro about extending methods.
 // could make it a composition but not going down that path yet
 type Override struct {
-	Path      string `json:"path"`
-	Value     any    `json:"value"`
-	Condition string `json:"condition"`
+	Path         string `json:"path"`
+	Value        any    `json:"value,omitempty"`
+	Condition    string `json:"condition"`
+	ValueProgram string `json:"valueProgram,omitempty"`
```
xiazhan: nit:  Would be Expression or ValueExpression better?

JinghanFu: I initially named it `ValueExpression` in `override.go` to reflect its `string` type, and `ValueProgram` in `mutation.go` to reflect its compiled `cel.Program` type. However, since the other three fields (`Path`, `Condition`, `Value`) use identical names across both files, I was concerned that having different names for this new field might introduce unnecessary confusion. So I opted to align them both as `ValueProgram` for consistency.

But if the naming discrepancy wouldn't cause confusion, I'm happy to rename it back to `ValueExpression` in `override.go` to more accurately reflect its string type!

docs/overrides.md
```diff
@@ -47,6 +47,29 @@ Supported fields:
...
+
+Use `valueProgram` instead of `value` to dynamically resolve the override value from the resource's current state using a CEL expression:
+
+```yaml
+annotations:
+  eno.azure.io/overrides: |
+    [
+      {
+        "path": "self.spec.resourcePolicy.containerPolicies[0].maxAllowed.memory",
+        "valueProgram": "self.spec.resourcePolicy.containerPolicies[0].maxAllowed.memory",
```
wonderyl: Rename here too?

internal/resource/mutation/mutation.go
```diff
@@ -24,26 +25,29 @@ var (
...
-	StatusMissingParent    Status = "MissingParent"
-	StatusIndexOutOfRange  Status = "IndexOutOfRange"
-	StatusPathTypeMismatch Status = "PathTypeMismatch"
+	StatusActive              Status = "Active"
+	StatusInactive            Status = "Inactive"
+	StatusInvalidCondition    Status = "InvalidCondition"
+	StatusMissingParent       Status = "MissingParent"
+	StatusIndexOutOfRange     Status = "IndexOutOfRange"
+	StatusPathTypeMismatch    Status = "PathTypeMismatch"
+	StatusInvalidValueProgram Status = "InvalidValueProgram"
```
wonderyl: Rename here too

internal/resource/mutation/mutation.go
```diff
@@ -90,7 +106,28 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
...
+	resolvedValue := o.Value
+	if o.ValueProgram != nil {
+		if current == nil {
+			logger.Info("skipping CEL value evaluation - current resource is nil", "path", o.Path.String())
+			return StatusInactive, nil
+		}
+		val, err := enocel.Eval(ctx, o.ValueProgram, comp, current, o.Path)
+		if err != nil {
+			logger.Error(err, "failed to evaluate value expression", "path", o.Path.String())
+			return StatusInvalidValueProgram, nil
```
wonderyl: should return error

internal/resource/mutation/mutation.go
```diff
@@ -90,7 +106,28 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
...
+		}
+		resolvedValue = val.Value()
+		if resolvedValue == nil || resolvedValue == structpb.NullValue_NULL_VALUE {
```
wonderyl: What if the user want to unset a value?

go.work.sum
```diff
@@ -530,11 +530,15 @@ cloud.google.com/go/workflows v1.12.3/go.mod h1:fmOUeeqEwPzIU81foMjTRQIdwQHADi/v
 cloud.google.com/go/workflows v1.12.4 h1:uHNmUiatTbPQ4H1pabwfzpfEYD4BBnqDHqMm2IesOh4=
 cloud.google.com/go/workflows v1.12.4/go.mod h1:yQ7HUqOkdJK4duVtMeBCAOPiN1ZF1E9pAMX51vpwB/w=
 cloud.google.com/go/workflows v1.13.0/go.mod h1:StCuY3jhBj1HYMjCPqZs7J0deQLHPhF6hDtzWJaVF+Y=
+codeberg.org/go-fonts/liberation v0.5.0/go.mod h1:zS/2e1354/mJ4pGzIIaEtm/59VFCFnYC7YV6YdGl5GU=
+codeberg.org/go-latex/latex v0.1.0/go.mod h1:LA0q/AyWIYrqVd+A9Upkgsb+IqPcmSTKc9Dny04MHMw=
```
ruinan-liu: Where are we using those changes?

internal/resource/mutation/mutation.go
```diff
@@ -90,10 +106,37 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
...
+		val, err := enocel.Eval(ctx, o.ValueExpression, comp, current, o.Path)
+		if err != nil {
+			logger.Error(err, "failed to evaluate value expression", "path", o.Path.String())
+			return StatusInvalidValueExpression, fmt.Errorf("evaluating value expression for path %s: %w", o.Path.String(), err)
+		}
+		resolvedValue = val.Value()
+
+		if resolvedValue == structpb.NullValue_NULL_VALUE {
+			logger.Info("value expression evaluated to null (explicit unset)", "path", o.Path.String())
+			return StatusActive, nil
```
ruinan-liu: nit: Why are we returning StatusActive here?

ruinan-liu (commented):
Minor comments

internal/resource/mutation/mutation.go
```diff
@@ -90,10 +106,37 @@ func (o *Op) Apply(ctx context.Context, comp *apiv1.Composition, current, mutate
...
+		}
```
xiazhan: Where do we apply the value as "null"?

xiazhan: Did we verify that it works in e2e? Shall we add an e2e test case as well?

PR #602, opened 2026-05-15
[eno] Add telemetry for MissingInput and OutofStepInputs

ruinan-liu:
Enhance telemetry so that we know what inputs are missing and what inputs are out of step

### DIFF
--- a/internal/controllers/composition/controller.go
+++ b/internal/controllers/composition/controller.go
@@ -265,7 +265,7 @@ func (c *compositionController) reconcileSimplifiedStatus(ctx context.Context, s
 	if next != nil && synth != nil {
 		switch next.Status {
 		case "MissingInputs":
-			logger.Info("composition is missing required inputs", "missingInputs", inputs.Missing(synth, comp))
+			logger.Info("composition is missing required inputs", "missingInputs", inputs.Missing(synth, comp), "expectedInputs", inputs.Expected(synth))
 		case "MismatchedInputs":
 			logger.Info("composition has inputs that are out of lockstep", "mismatchedInputs", inputs.Mismatched(synth, comp, comp.Status.InputRevisions), "synthesizerGeneration", synth.Generation, "compositionGeneration", comp.Generation)
 		}
--- a/internal/controllers/composition/controller.go
+++ b/internal/controllers/composition/controller.go
@@ -33,6 +33,8 @@ const (
 	AKSComponentLabel                   = "aks.azure.com/component-type" // TODO(ruinanliu): Temp workaround remove after 14802391 is released
 	addOnLabelValue                     = "addon"                        // TODO(ruinanliu):  Temp workaround remove after 14802391 is released
 	EnoCleanupFinalizer                 = "eno.azure.io/cleanup"
+	MissingInputStatus                  = "MissingInputs"
+	MismatchedInputsStatus              = "MismatchedInputs"
 )
 
 type compositionController struct {
@@ -262,11 +264,11 @@ func (c *compositionController) reconcileSimplifiedStatus(ctx context.Context, s
 		return false, nil
 	}
 
-	if next != nil && synth != nil {
+	if synth != nil {
 		switch next.Status {
-		case "MissingInputs":
+		case MissingInputStatus:
 			logger.Info("composition is missing required inputs", "missingInputs", inputs.Missing(synth, comp), "expectedInputs", inputs.Expected(synth))
-		case "MismatchedInputs":
+		case MismatchedInputsStatus:
 			logger.Info("composition has inputs that are out of lockstep", "mismatchedInputs", inputs.Mismatched(synth, comp, comp.Status.InputRevisions), "synthesizerGeneration", synth.Generation, "compositionGeneration", comp.Generation)
 		}
 	}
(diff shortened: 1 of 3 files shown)


internal/controllers/composition/controller.go
```diff
@@ -262,6 +262,15 @@ func (c *compositionController) reconcileSimplifiedStatus(ctx context.Context, s
 		return false, nil
 	}
 
+	if next != nil && synth != nil {
+		switch next.Status {
+		case "MissingInputs":
```
tilnl: nit: extracting into const?

tilnl (approved):

internal/inputs/inputs.go
```diff
@@ -19,44 +26,75 @@ func Exist(syn *apiv1.Synthesizer, c *apiv1.Composition) bool {
...
+
+// MismatchedInput describes a single input revision that is out of lockstep with the others
+// (or with the current synthesizer/composition generation). It is intended for logging/telemetry.
+type MismatchedInput struct {
+	Key                   string
+	Revision              *int
+	MaxRevision           *int
+	SynthesizerGeneration *int64
+	CompositionGeneration *int64
+}
```
xiazhan, line 62: **Pointer rendering in logs.** `MismatchedInput` uses `*int` / `*int64`. Depending on the logr/zap encoder in use, pointer fields can render as memory addresses (`0xc0000...`) rather than dereferenced values. Worth verifying against the production logging config — or adding a `String()` / `MarshalLog` on `MismatchedInput`, or switching to value types with a sentinel — so logs are always human-readable.

internal/inputs/inputs.go
```diff
@@ -19,44 +26,75 @@ func Exist(syn *apiv1.Synthesizer, c *apiv1.Composition) bool {
...
-		if rev.Revision != nil && *maxRevision != *rev.Revision {
-			return true
+		stale := false
+		if rev.SynthesizerGeneration != nil && *rev.SynthesizerGeneration < synth.Generation {
+			stale = true
+		}
+		if rev.CompositionGeneration != nil && *rev.CompositionGeneration < comp.Generation {
+			stale = true
 		}
+		revMismatch := maxRevision != nil && rev.Revision != nil && *rev.Revision != *maxRevision
```
xiazhan: **Behavioral equivalence note.** The old `OutOfLockstep` short-circuited on the first stale generation before computing `maxRevision`; this version walks all revs first, then flags per-key. Boolean output via `OutOfLockstep` is preserved, but stale-only entries can now have `MaxRevision` populated when *other* entries declare revisions (covered by the `Stale gens and revision mismatch combined` test). Just confirm that shape is what the telemetry consumers expect.

internal/controllers/composition/controller.go
```diff
@@ -262,6 +262,15 @@ func (c *compositionController) reconcileSimplifiedStatus(ctx context.Context, s
...
+			logger.Info("composition is missing required inputs", "missingInputs", inputs.Missing(synth, comp), "expectedInputs", inputs.Expected(synth))
+		case "MismatchedInputs":
+			logger.Info("composition has inputs that are out of lockstep", "mismatchedInputs", inputs.Mismatched(synth, comp, comp.Status.InputRevisions), "synthesizerGeneration", synth.Generation, "compositionGeneration", comp.Generation)
```
xiazhan, line 272: **Duplicate evaluation.** `Missing`/`Mismatched` are recomputed here even though `buildSimplifiedStatus` already evaluated equivalent conditions to set `next.Status`. Cheap, but having `buildSimplifiedStatus` return the diagnostic data would avoid the double traversal.\n\nAlso minor: `MissingInputs` / `MismatchedInputs` are repeated string literals across the file — promoting them to typed constants would prevent drift (out of scope here). And the `next != nil && synth != nil` guard on line 265 — if `next` cannot be nil at this point, the guard is dead (harmless either way).

xiazhan (commented):
Code review — LGTM with non-blocking comments. Nice refactor: `Exist`/`OutOfLockstep` delegating to richer `Missing`/`Mismatched` helpers is a clean way to surface telemetry, and the new tests are thorough. A few inline notes below.

internal/controllers/composition/controller.go
ruinan-liu: Done, thanks changed `MissingInputs` and `MismatchedInputs` to use constant values and removed next != nil deadcode.

I don't think having `buildSimplifiedStatus` return the diagnostic data would be helpful, because this entangles a pure status-derivation function and this would change a lot of the functions. The diagnostic computation only run on the status-change path, so this isn't on the hot path. `Inputs.Missing` is `O(refs X revs)`, `input.Mismatched` is `O(revs)`. Synthesizers typically declare a handful of refs, the constant factor is negligible.

internal/inputs/inputs.go
ruinan-liu: Yes, the data-shape difference you noted, the stale-only entries getting `MaxRevisions` populated when peers declare revieisons is intentional. `MaxRevision` is documented as contextual (The max revision across all inputs), not as an accusation about the flagged entry. Populating it unifomly means a single log line is self-contained and the operation can read it without cross-referncing other entries. THe only consumer today is the log line `"composition has inputs that are out of lockstep"` in `reconcileSimplifiedStatus. No metrics, no external parsers, no status field, so there are no compatibility surface to worry about.

internal/inputs/inputs.go
ruinan-liu: Switched to `mismatchedInput` to value types `int/int64` with 0 as the unset sentinel, k8s generations and revisions are always >= 1 in practice, so 0 is unambigious and the struct now renders clearnly under any encoder without needing a custom `string()/MarshalLog`. The `mismatched` builder only copies dereferenced values when the soruce `InputRevision` pointer is non-nil, so absent fields stay at 0 in the log output. 

Also confirmed in production logger uses zap's JSON encoder (via zapr), which would have dereferenced pointers via `encoding/json` anyway but value types are simpler and remove the encoder dependency from the contrcact.

internal/controllers/composition/controller.go
ruinan-liu: DOne, thanks

xiazhan (approved):

PR #618, opened 2026-06-30
[eno] Fast track cancellation for out of step synthesis pods

ruinan-liu:
Instead of waiting for 2 minutes straight, we want to fast track cancellation for out of lock synthesis pods.

### DIFF
--- a/internal/controllers/composition/controller.go
+++ b/internal/controllers/composition/controller.go
@@ -191,7 +191,7 @@ func (c *compositionController) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 
 		if terminal {
-			if grace := time.Until(pod.CreationTimestamp.Time.Add(podCompletionGracePeriod)); grace > 0 {
+			if grace := time.Until(podTerminationTime(pod).Add(podCompletionGracePeriod)); grace > 0 {
 				return ctrl.Result{RequeueAfter: grace}, nil
 			}
 
@@ -640,4 +640,22 @@ func (c *compositionController) terminalInFlightPod(ctx context.Context, comp *a
 		}
 	}
 	return nil, false, nil
+}
+
+// podTerminationTime returns the time the pod actually finished running: the
+// latest container termination timestamp. This is the correct anchor for the
+// completion grace period ("how long after the pod finished do we wait"). It
+// falls back to CreationTimestamp when no terminated state is recorded, which
+// shouldn't happen for a pod already observed in a terminal phase.
+func podTerminationTime(pod *corev1.Pod) time.Time {
+	var latest time.Time
+	for _, cs := range pod.Status.ContainerStatuses {
+		if term := cs.State.Terminated; term != nil && term.FinishedAt.Time.After(latest) {
+			latest = term.FinishedAt.Time
+		}
+	}
+	if latest.IsZero() {
+		return pod.CreationTimestamp.Time
+	}
+	return latest
 }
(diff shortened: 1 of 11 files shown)


internal/controllers/composition/controller.go
```diff
@@ -39,17 +40,32 @@ const (
 	NotReadyStatus                      = "NotReady"
 	ReconcilingStatus                   = "Reconciling"
 	LogSampleCap                        = 50
+	synthesisIDLabelKey                 = "eno.azure.io/synthesis-uuid"
```
tilnl: Nit: This duplicates the unexported synthesisIDLabelKey in internal/controllers/synthesis/pod.go. Not a blocker but consider to keep a single source of truth.

internal/controllers/composition/controller.go
```diff
@@ -556,3 +615,47 @@ func (c *compositionController) clearDependencyStatus(ctx context.Context, comp
...
+	err := c.client.List(ctx, &pods,
+		client.InNamespace(c.synthesisPodNamespace),
+		client.MatchingLabels{synthesisIDLabelKey: syn.UUID})
+	if err != nil {
+		return nil, false, fmt.Errorf("error listing synthesizer pods: %w", err)
+	}
+
+	for i := range pods.Items {
+		pod := &pods.Items[i]
+		if pod.DeletionTimestamp != nil {
```
xiazhan, line 634: Does it mean garbage collector already detect the pod's terminal state when deletionTimeStamp is not nil?

internal/controllers/composition/controller.go
```diff
@@ -556,3 +615,47 @@ func (c *compositionController) clearDependencyStatus(ctx context.Context, comp
...
+			continue
+		}
+		if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed {
```
xiazhan: Failed pod will be restarted by k8s given that the RestartPolicy is OnFailure. Are we sure that we'll cancel it?

internal/controllers/composition/controller.go
ruinan-liu: Done, thanks for pointing this out.

internal/controllers/composition/controller.go
ruinan-liu: gc. go is what manages these pods' lifecycle — it deletes a pod (with a reason) once the pod's state change and the Composition's synthesis status line up (canceled → timeout, UUID mismatch → superseded, in-flight swapped to current → success). So when DeletionTimestamp != nil, it means gc.go has already issued the delete for that pod, and there's nothing more for the fast-cancel path to do with it — we skip it and let the genuine terminal-phase check (PodSucceeded || PodFailed) below handle pods that are still present.

internal/controllers/composition/controller.go
ruinan-liu: Dropped, the podFailed case. The case where it does nothing will output pod succeeded
if reason, skip := skipSynthesis(comp, syn, env); skip {
		logger.Info("synthesis is no longer relevant - skipping", "reason", reason)
		return nil
	}

ruinan-liu (commented):
Replied to comments

internal/controllers/composition/controller.go
```diff
@@ -556,3 +614,47 @@ func (c *compositionController) clearDependencyStatus(ctx context.Context, comp
 	}
 	return ctrl.Result{}, nil
 }
+
+func (c *compositionController) terminalInFlightPod(ctx context.Context, comp *apiv1.Composition) (*corev1.Pod, bool, error) {
```
xiazhan: rename to succeededInfFlightPod?

ruinan-liu: Done, thanks

ruinan-liu (commented):
Reply to comment

xiazhan (approved):

PR #619, opened 2026-07-01
[eno] Add default reconcile interval support

ruinan-liu:
## Summary

Adds an opt-in **default reconcile interval** for resources managed by Eno.

When a Composition (or an individual resource) is annotated with
`eno.azure.io/use-default-reconcile-interval: "true"`, every managed resource
that does **not** set an explicit `eno.azure.io/reconcile-interval` will be
periodically requeued using the reconciler's default interval (**15s**,
configurable via `--default-reconcile-interval`).

This lets us enable periodic drift correction for an entire Composition with a
single annotation, without asking partner teams to annotate each of their own
manifests.

## Motivation

Without a reconcile interval, Eno only re-applies a resource on the next
synthesis. If a customer-visible resource is deleted or modified out-of-band, it
can stay drifted until then. Opting into the default interval ensures such drift
(e.g. an accidental deletion) is corrected promptly.

## Behavior

Interval precedence when deciding whether/when to requeue:

1. Explicit `eno.azure.io/reconcile-interval` — always wins.
2. Else, if `eno.azure.io/use-default-reconcile-interval: "true"` — use the
   default interval.
3. Else — no periodic requeue (unchanged behavior).

The resolved interval is still floored by `--min-reconcile-interval`. The
`use-default` annotation cascades from the Composition to its resources, matching
existing annotations like `disable-reconciliation`.

## Changes

- New `--default-reconcile-interval` flag (default `15s`) on `eno-reconciler`.
- `use-default-reconcile-interval` opt-in plumbed through the resource snapshot
  and honored in the reconciler's requeue logic.
- Backward compatible: resources with neither annotation are unaffected.

## Testing

- Unit: requeue interval-resolution matrix (explicit wins, default opt-in,
  min-interval flooring, no-op when not opted in).
- Resource: annotation parsing for the new opt-in.
- Integration: out-of-Eno modification converges via the default interval.
- E2E: out-of-band deletion of ServiceAccount/ClusterRole/RoleBinding/Deployment
  is recreated purely from the periodic requeue (verified via new UIDs).

### DIFF
--- a/cmd/eno-reconciler/main.go
+++ b/cmd/eno-reconciler/main.go
@@ -57,7 +57,7 @@ func run() error {
 	flag.DurationVar(&recOpts.Timeout, "timeout", time.Minute, "Per-resource reconciliation timeout. Avoids cases where client retries/timeouts are configured poorly and the loop gets blocked")
 	flag.DurationVar(&recOpts.ReadinessPollInterval, "readiness-poll-interval", time.Second*5, "Interval at which non-ready resources will be checked for readiness")
 	flag.DurationVar(&recOpts.MinReconcileInterval, "min-reconcile-interval", time.Second, "Minimum value of eno.azure.com/reconcile-interval that will be honored by the controller")
-	flag.DurationVar(&recOpts.DefaultReconcileInterval, "default-reconcile-interval", time.Second*15, "Reconcile interval used for resources that set 'eno.azure.io/use-default-reconcile-interval: true' but don't specify an explicit 'eno.azure.io/reconcile-interval'")
+	flag.DurationVar(&recOpts.DefaultReconcileInterval, "default-reconcile-interval", time.Minute*5, "Reconcile interval used for resources that set 'eno.azure.io/use-default-reconcile-interval: true' but don't specify an explicit 'eno.azure.io/reconcile-interval'")
 	flag.BoolVar(&recOpts.DisableServerSideApply, "disable-ssa", false, "Use non-strategic three-way merge patches instead of server-side apply")
 	flag.StringVar(&compositionSelector, "composition-label-selector", labels.Everything().String(), "Optional label selector for compositions to be reconciled")
 	flag.StringVar(&compositionNamespace, "composition-namespace", metav1.NamespaceAll, "Optional namespace to limit compositions that will be reconciled")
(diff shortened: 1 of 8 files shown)


xiazhan:
Just want make sure that it's not against the maintenance configuration. If a customer set a maintenance configuration, they won't expect a change. The composition seems to still keep unchanged even with this change, thus it should be fine.

ruinan-liu:
@xiazhan I think this should be fine, it is an annotation that we need to set from overlaymgr. The composition will remain unchanged. I believe maintenance windows is being set in overlaymgr, so if overlaymgr did not initiate any new operation then eno-reconciler will not initiate any reconciliation. 

The main use case is for us to add this to addon resources, this way if cx accidently deletes a resource we can reconcile it back without customer needing to trigger `az aks update` or other things.

tilnl:
15s seems to be quite short interval. For addon, i think 5m should be enough.

tilnl:
> @xiazhan I think this should be fine, it is an annotation that we need to set from overlaymgr. The composition will remain unchanged. I believe maintenance windows is being set in overlaymgr, so if overlaymgr did not initiate any new operation then eno-reconciler will not initiate any reconciliation.
> 
> The main use case is for us to add this to addon resources, this way if cx accidently deletes a resource we can reconcile it back without customer needing to trigger `az aks update` or other things.

This default interval reconciliation  only trigger reconcile resourceslice to expect state, not a resynthesis, right?

ruinan-liu:
> > @xiazhan I think this should be fine, it is an annotation that we need to set from overlaymgr. The composition will remain unchanged. I believe maintenance windows is being set in overlaymgr, so if overlaymgr did not initiate any new operation then eno-reconciler will not initiate any reconciliation.
> > The main use case is for us to add this to addon resources, this way if cx accidently deletes a resource we can reconcile it back without customer needing to trigger `az aks update` or other things.
> 
> This default interval reconciliation only trigger reconcile resourceslice to expect state, not a resynthesis, right?

Yes, not a resynthesis just trigger the resource to be back to the expected state.

ruinan-liu:
> 15s seems to be quite short interval. For addon, i think 5m should be enough.

Changed to 5 minutes by default

tilnl (approved):

PR #639, opened 2026-08-20
[eno] Allow maxConcurrentReconciles for Symphony/Composition/input watch controllers

ruinan-liu:
Make them tunable

### DIFF
--- a/internal/flowcontrol/inputrevbuffer.go
+++ b/internal/flowcontrol/inputrevbuffer.go
@@ -2,6 +2,7 @@ package flowcontrol
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"sync"
 	"time"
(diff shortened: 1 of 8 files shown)


internal/flowcontrol/inputrevbuffer.go
```diff
@@ -109,8 +126,17 @@ func (w *CompositionInputRevisionWriteBuffer) Start(ctx context.Context) error {
...
 		w.queue.ShutDown()
 	}()
-	for w.processQueueItem(ctx) {
+
+	var workers sync.WaitGroup
+	workers.Add(w.workers)
+	for range w.workers {
+		go func() {
+			defer workers.Done()
+			for w.processQueueItem(ctx) {
```
xiazhan, line 136: Is it thread safe?

ruinan-liu: Yes, it it safe for the intended lifecycle. We have multiple gorountines consume one concurrent-safe workqueue. we use `mut.lock` to protect the shared maps. 

Input revisions are also deep copied before async processing, preventing caller-owned data races. K8s also uses optimistic locking and retry on conflicts. Have also added concurrent tests in inputrevbuffer_test.go

ruinan-liu (commented):
REspond to comments

tilnl (approved):

PR #647, opened 2026-09-04
[eno] Fix readiness handling for deleted tombstone resource

ruinan-liu:
ENO continued evaluating normal readiness expressions after tombstoned resources had been successfully deleted. Because readiness expressions cannot pass when the target object no longer exists, the ResourceState remained `Deleted=true` and `Reconciled=true` but `Ready=nil`, leaving the Composition stuck in `NotReady`.

This change identifies deletion snapshots before evaluating normal object readiness. Once deletion is confirmed, ENO treats deletion completion as the tombstone's terminal readiness state. This preserves existing readiness and deletion ordering while allowing the Composition to become ready.

Added unit, integration, ordering, ResourceSlice aggregation, and live-cluster E2E coverage for removing a readiness-enabled Service in a subsequent synthesis

### DIFF
--- a/internal/controllers/reconciliation/controller.go
+++ b/internal/controllers/reconciliation/controller.go
@@ -192,7 +192,7 @@ func (c *Controller) Reconcile(ctx context.Context, req resource.Request) (ctrl.
 	deleted := markResourceAsDeleted(current, snap, failingOpen)
 
 	// For tombstones, confirmed deletion is the terminal readiness state. Explicit readiness checks cannot be evaluated after the target object has disappeared
-	if snap.Deleted() && deleted && ready == nil {
+	if snap != nil && snap.Deleted() && deleted && ready == nil {
 		now := metav1.Now()
 		ready = &now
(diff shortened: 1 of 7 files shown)


internal/controllers/reconciliation/controller.go
```diff
@@ -213,31 +220,39 @@ func (c *Controller) reconcileResource(ctx context.Context, comp *apiv1.Composit
...
 	logger.Info("creating resource snapshot")
 	snap, err = resource.Snapshot(ctx, comp, current)
 	if err != nil {
 		logger.Error(err, "failed to create resource snapshot")
 		return nil, nil, nil, false, fmt.Errorf("failed to create resource snapshot: %w", err)
 	}
+
+	status := resource.State()
+	if snap.Deleted() {
+		// The desired state is deletion. Normal object readiness expressions should no longer apply
```
tilnl: so we mark status ready when it is supposed to be deleted?

xiazhan (approved):

PR #651, opened 2026-09-10
[eno][part1/2] Continue synthesis with incomplete ResourceSlice history

ruinan-liu:
Allow synthesis to proceed when historical ResourceSlice references are nil, empty, or point to missing objects, instead of blocking on unavailable history.

• Generate desired resources and tombstones using the historical slices that remain available.
• Set  TombstoneRecoveryRequired  when history is incomplete or no current synthesis exists, and carry the flag forward across subsequent syntheses.
• Allow reconstitution to skip unavailable previous history while preserving strict loading of current desired resources and existing handling of non-404 API errors.
• Handle malformed references safely in resynthesis requests, cleanup, and not-ready diagnostics.
• Include the recovery flag in the Composition CRD schema.

This is Part 1 only: the flag does not block synthesis or reconciliation. Recovering deletions from unavailable history and clearing the flag are deferred to a later change.

Testing

Added 36 test groups with 170 subtests covering unit, executor-functional, and envtest controller-integration scenarios. All new tests pass, including race checks.


internal/controllers/resourceslice/slice.go
```diff
@@ -50,6 +51,13 @@ func (s *sliceController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	snapshot := statusSnapshot{Reconciled: true, Ready: true}
 
 	for _, ref := range comp.Status.CurrentSynthesis.ResourceSlices {
+		if ref == nil || ref.Name == "" {
```
xiazhan: Is it the only sign which indicate the corruption of ResourceSlice?

internal/execution/executor.go
```diff
@@ -85,7 +87,7 @@ func (e *Executor) Synthesize(ctx context.Context, env *Env) error {
 			Severity: krmv1.ResultSeverityError,
 		}}}
 
-		if err := e.updateComposition(ctx, env, comp, syn, sliceRefs, revs, output); err != nil {
+		if err := e.updateComposition(ctx, env, comp, syn, sliceRefs, recoveryRequired, revs, output); err != nil {
```
xiazhan, line 90: recoveryRequired is always false here? Maybe just pass in "false".

internal/controllers/reconciliation/reconstitution.go
```diff
@@ -120,43 +120,78 @@ func (r *reconstitutionSource) Reconcile(ctx context.Context, req ctrl.Request)
...
+	logger := logr.FromContextOrDiscard(ctx).WithValues(
+		"synthesisUUID", synthesis.UUID,
+		"isPreviousSynthesis", isPreviousSynthesis,
+	)
+
+	cachedSlices := make([]apiv1.ResourceSlice, 0, len(synthesis.ResourceSlices))
 	for i, ref := range synthesis.ResourceSlices {
+		if ref == nil || ref.Name == "" {
+			if !isPreviousSynthesis {
+				logger.V(1).Info("current synthesis slice reference has no name; waiting for resynthesis", "referenceIndex", i)
```
xiazhan: This log can be misleading because it also includes the case in which the resourceSlice is nil.

internal/controllers/reconciliation/reconstitution.go
```diff
@@ -120,43 +120,78 @@ func (r *reconstitutionSource) Reconcile(ctx context.Context, req ctrl.Request)
...
+				return false, nil
+			}
+
+			logger.V(1).Info("previous synthesis slice reference has no name; skipping", "referenceIndex", i)
```
xiazhan: The same here.

internal/execution/executor.go
ruinan-liu: Done, thanks

ruinan-liu (commented):
Reply to comment

internal/controllers/reconciliation/reconstitution.go
ruinan-liu: Updated the telemetry

internal/controllers/reconciliation/reconstitution.go
ruinan-liu: Update telemetry

ruinan-liu (commented):
Reply to comment

xiazhan (approved):

PR #654, opened 2026-09-17
[eno][backup 2/5] Add inventory format and tombstone diff helpers

ruinan-liu:
Add the inventory format and missing-tombstone helpers for the backup operator. This PR contains helpers only: it does not register a controller, persist Kubernetes objects, or gate reconstitution. Controller integration remains in Part 2; deletion-event retirement and periodic inventory GC remain separate follow-ups.

## Inventory format

Each inventory is a ConfigMap in `kube-system`, named `eno-inventory-<lineageHash>-<synthesisUUID>` and labeled `eno.azure.io/inventory-lineage=<lineageHash>`. The lineage is derived from the Composition namespace and synthesizer name, excluding the Composition UID so inventories can survive Composition recreation.

`data["inventory.json"]` contains only a JSON array of resource identities, labels, and the `eno.azure.io/readiness-group` / `eno.azure.io/deletion-group` annotations. Empty inventories contain `[]`. Format version, Composition namespace, synthesizer name, synthesis UUID, and source synthesis timestamp are stored in ConfigMap annotations. No saved Composition or Symphony metadata objects, inventory wrapper type, or compatibility with the old JSON-envelope layout are included.

## Main functions

- `makeInventory`: Converts the supplied ResourceSlices directly into a `*corev1.ConfigMap`. Skips deleted manifests and Eno Patch pseudo-resources, keeps the first occurrence of each group/kind/namespace/name identity, normalizes and sorts once, and enforces the 1 MiB inventory data limit.
- `inventoryLineage` / `inventoryName`: Derive the stable namespace/synthesizer lineage and the per-synthesis ConfigMap name.
- `decodeInventorySnapshot`: Decodes `inventory.json` into `[]inventoryResource`, checks the ConfigMap namespace and deletion state, and validates identity agreement. Does not sort, normalize, or repeat the write-side size check.
- `validateInventoryIdentity`: Checks the source namespace and synthesizer against the supplied Composition, the lineage label, and agreement between the ConfigMap name and its synthesis UUID annotation. Historical synthesis UUIDs remain valid recovery inputs.
- `selectInventory`: Returns a deep copy of the ConfigMap with the greatest source synthesis timestamp. Equal timestamps keep the first candidate; empty input returns nil. Reports invalid timestamps without decoding resource payloads or rechecking lineage.
- `inventoriesMatch`: Compares decoded resources in their stored order, the snapshot metadata annotations, and source timestamps. Reports decoding and timestamp errors without repeating normalization.
- `missingTombstones`: Compares inventory identities against current ResourceSlices and returns `Deleted: true` manifests for missing identities. Existing full resource manifests and tombstones prevent duplicate tombstones; API version is not part of the identity key. Preserves saved labels and group annotations and skips Patches rather than resolving their targets.

## Controller contract

All resources within a Composition are handled by one eno-reconciler. The controller is responsible for selecting the Composition using the existing `--resource-filter` and supplying its ResourceSlices and initialized current synthesis. These helpers perform no CEL filtering or partial Composition evaluation.

Patches are excluded from both inventory construction and the recovery comparison: a Patch alone does not count as its target's full resource manifest in `missingTombstones`.


internal/controllers/backup/inventory.go
```diff
@@ -0,0 +1,323 @@
...
+		return fmt.Errorf("label %s=%q does not match lineage %q", inventoryLineageLabel, item.Labels[inventoryLineageLabel], inventoryLineage(comp))
+	}
+	if name := inventoryName(comp, item.Annotations[inventorySynthesisUUIDAnnotation]); item.Name != name {
+		return fmt.Errorf("name %q must be %q", item.Name, name)
+	}
+	return nil
+}
+
+func makeInventory(comp *apiv1.Composition, slices []apiv1.ResourceSlice) (*corev1.ConfigMap, error) {
+	syn := comp.Status.CurrentSynthesis
```
tilnl: Missing nil guard on `comp.Status.CurrentSynthesis`. Both `syn.UUID` and `syn.Synthesized.UTC()` panic if the current synthesis is nil or `Synthesized` is unset — and `Synthesized` is legitimately nil for an in-flight synthesis. The PR description says the caller guarantees an initialized ready synthesis, but a panic in a controller reconcile loop is an expensive way to enforce a contract that lives only in a doc. Suggest returning an error:

```go
syn := comp.Status.CurrentSynthesis
if syn == nil || syn.Synthesized == nil {
    return nil, fmt.Errorf("composition has no completed synthesis")
}
```

internal/controllers/backup/inventory.go
```diff
@@ -0,0 +1,323 @@
...
+				inventorySynthesisUUIDAnnotation:        syn.UUID,
+				inventorySynthesizedAnnotation:          syn.Synthesized.UTC().Format(time.RFC3339),
+			},
+		},
+		Data: map[string]string{inventoryDataKey: string(payload)},
+	}, nil
+}
+
+func missingTombstones(slices []apiv1.ResourceSlice, resources []inventoryResource) ([]apiv1.Manifest, error) {
+	if resources == nil {
```
tilnl: Distinguishing `nil` from empty here is fragile as a safety gate. `decodeInventorySnapshot` returns whatever `json.Unmarshal` produces, and a payload of literal `null` decodes to a nil slice that passes every validation check — so a corrupted/hand-edited inventory silently yields zero tombstones instead of being rejected. Recommend rejecting `null` explicitly in `decodeInventorySnapshot` (`if data == nil { return nil, fmt.Errorf("inventory.json must be a JSON array") }`) so this guard only ever sees a real absence-of-inventory signal from the caller.

internal/controllers/backup/inventory.go
```diff
@@ -0,0 +1,323 @@
...
+		return nil, nil
+	}
+	existing := map[resource.Ref]struct{}{}
+	for _, slice := range slices {
+		for i, manifest := range slice.Spec.Resources {
+			_, res, err := parseInventoryManifest(manifest.Manifest)
+			if err != nil {
```
tilnl: Inconsistent handling of unparseable manifests between the two helpers: `makeInventory` skips `manifest.Deleted` *before* parsing, so a corrupt tombstone manifest is tolerated on the write path, while this loop parses every manifest including deleted ones and aborts the entire recovery with an error. The existing test fixture (`{Manifest: "{", Deleted: true}`) passes `makeInventory` but would fail here.

Since deleted manifests only need their identity, and a corrupt one cannot contribute one, consider either skipping parse failures on deleted entries or — better — parsing deleted entries in `makeInventory` too so both paths agree on what constitutes a valid slice.

internal/controllers/backup/inventory.go
```diff
@@ -0,0 +1,323 @@
...
+			seen[ref] = struct{}{}
+			data = append(data, res)
+		}
+	}
+	normalizeInventoryResources(data)
+	payload, err := json.Marshal(data)
+	if err != nil {
+		return nil, fmt.Errorf("encoding inventory: %w", err)
+	}
+	if len(payload) > inventoryMaxDataBytes {
```
tilnl: This only bounds the `inventory.json` payload, but the 1 MiB API-server limit applies to the sum of all `data` + `binaryData` entries, and the etcd object limit (~1.5 MiB) additionally covers annotations, labels, and managed fields. A payload at exactly `inventoryMaxDataBytes` plus annotations will be rejected by the API server at `Create` time with an opaque error. Suggest reserving headroom (e.g. compare against `inventoryMaxDataBytes - 4096`) or computing the limit against the marshalled ConfigMap rather than the payload alone.

internal/controllers/backup/inventory.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	}
+	return obj, res, nil
+}
+
+func (res inventoryResource) ref() resource.Ref {
+	return resource.Ref{Group: res.Group, Kind: res.Kind, Namespace: res.Namespace, Name: res.Name}
+}
+
+func (res inventoryResource) isPatch() bool {
```
tilnl: Pinning `Version == "v1"` means a Patch served under any future `eno.azure.io` version stops being recognized as a pseudo-resource: it would be recorded in the inventory and later resurrected as a real tombstone targeting a nonexistent `Patch` object. Matching on group+kind only is safer here:

```go
return res.Group == "eno.azure.io" && res.Kind == "Patch"
```

internal/controllers/backup/inventory.go
tilnl, line 175: Missing nil guard on `comp.Status.CurrentSynthesis`. Both `syn.UUID` and `syn.Synthesized.UTC()` are dereferenced unconditionally, and `Synthesized` is legitimately nil for an in-flight synthesis. The caller contract is documented in the PR description but not enforced in code — a panic in a reconcile loop is an expensive way to discover it once Part 2 wires this up.

```go
syn := comp.Status.CurrentSynthesis
if syn == nil || syn.Synthesized == nil {
    return nil, fmt.Errorf("composition has no completed synthesis")
}
```

Non-blocking — fine to fold into the controller PR if the error shape is clearer there.

tilnl (approved):
Approving — helpers only, nothing calls them yet, and the test coverage is unusually thorough (tie-breaking, timezone-offset instants vs string compare, deep-copy independence, ordering determinism, old-envelope rejection).

Two non-blocking inline comments: a missing nil guard on `CurrentSynthesis` in `makeInventory`, and an asymmetry in how `makeInventory` and `missingTombstones` treat unparseable deleted manifests. Both are fine to fold into the controller PR.

Design choices I liked: the lineage hash deliberately excluding the Composition UID so inventories survive recreation, selection keyed on source synthesis timestamp rather than ConfigMap creation time, and `missingTombstones` building its `existing` set from tombstones as well as live manifests so it is idempotent by construction.

internal/controllers/backup/inventory.go
ruinan-liu: Thanks Tignting, yes in Part 3 we have this guard
syn := comp.Status.CurrentSynthesis
if syn == nil || syn.Synthesized == nil {
    return ctrl.Result{}, nil
}

ruinan-liu (commented):
Resolve comments

